PR #16, opened 2015-11-07
Add test for checking value type

alensuljkanovic:
Test raises EOF exception when instead of INT value, you set FLOAT.
Exception should be raised, but the message should indicate that
parser expects INT instead some other value.


igordejanovic:
Fixed in Arpeggio.

PR #47, opened 2017-11-23
FLOAT terminal now matches floats ending with "."

### LINKED ISSUE #37: Parsing of floats ending with period.
`FLOAT` seems to parse both `2.0` and `2`, but not `2.`, is this intended behavior?

borismarin:
Small adaptation to FLOAT regex to be able to cope with cases such as `1.e-2` . Closes #37.

### DIFF
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -89,7 +89,7 @@ def comment_block():        return _(r'/\*(.|\n)*?\*/')
 ID          = _(r'[^\d\W]\w*\b', rule_name='ID', root=True)
 BOOL        = _(r'(True|true|False|false|0|1)\b', rule_name='BOOL', root=True)
 INT         = _(r'[-+]?[0-9]+\b', rule_name='INT', root=True)
-FLOAT       = _(r'[+-]?(\d+(\.\d*)?|\.\b\d+)([eE][+-]?\d+)?(?<=[\w\.])(?![\w\.])',
+FLOAT       = _(r'[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?(?<=[\w\.])(?![\w\.])',
                 'FLOAT', root=True)
 STRING      = _(r'("(\\"|[^"])*")|(\'(\\\'|[^\'])*\')', 'STRING', root=True)
 NUMBER      = OrderedChoice(nodes=[FLOAT, INT], rule_name='NUMBER', root=True)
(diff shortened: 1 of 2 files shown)


borismarin:
@igordejanovic the build below is failing due to travis [not finding python 3.4](https://travis-ci.org/igordejanovic/textX/builds/306405528), not because of the code...

igordejanovic:
Thanks for the contribution, Boris. I would like to run some additional testing during the weekend as this touches fundamental parts and could make a lot of trouble if done wrong.
Yeah, I noticed that Travis breaks a lot on not finding Python 3.4.

borismarin:
@igordejanovic it is indeed tricky - for example, this new regex now matches strings like `e1`, which is probably unintended.

textx/lang.py
```diff
@@ -89,7 +89,7 @@ def comment_block():        return _(r'/\*(.|\n)*?\*/')
 ID          = _(r'[^\d\W]\w*\b', rule_name='ID', root=True)
 BOOL        = _(r'(True|true|False|false|0|1)\b', rule_name='BOOL', root=True)
 INT         = _(r'[-+]?[0-9]+\b', rule_name='INT', root=True)
-FLOAT       = _(r'[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?\b', 'FLOAT',
+FLOAT       = _(r'[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?', 'FLOAT',
```
igordejanovic: What's the reason behind removing `\b` at the end? There are some subtle gotchas if anchoring is not used. I made an extension to float test in cf22658 that passes with original textX but fails with this extension. The problem is when you have something that matches FLOAT as a prefix of some other match.

igordejanovic (changes requested):
Looks good to me. The only issue is with anchoring. Please see the comment.

borismarin:
@igordejanovic , a simple `\b` would not be sufficient (at least given my level of regex wizardry) as the literal `.` doesn't count as a word boundary - I've added a custom boundary that behaves just like `\b`  (looks for a word char /not word char transition) but includes `.` as a word character.

textx/lang.py
borismarin: a simple \b would not be sufficient as the literal `.` doesn't count as a word boundary - I've added a custom boundary that behaves just like \b but looks for a word char /not word char transition including `.` as a word character.

igordejanovic:
Looks good. Thanks!

PR #52, opened 2018-01-12
Added cross-references list and rules dictionary fields to model

danixeee:
Added features are used by [textX-languageserver](https://github.com/textX-tools/textX-languageserver) for better performance.


igordejanovic:
@danixeee Tests are failing. Can you investigate?

textx/model.py
```diff
@@ -450,6 +467,32 @@ def _inner_resolve_link_rule_ref(cls, obj_name):
             result = _inner_resolve_link_rule_ref(obj_ref.cls,
                                                   obj_ref.obj_name)
             if result:
+                class RefRulePosition(object):
```
igordejanovic: This class should be moved out of this place. No need to create it multiple times, and the readability suffers.

textx/exceptions.py
```diff
@@ -1,14 +1,22 @@
...
+    def __init__(self, message, line=None, col=None, err_type=None):
         super(TextXError, self).__init__(message.encode('utf-8'))
         self.line = line
         self.col = col
+        self.err_type = err_type
 
 
 class TextXSemanticError(TextXError):
-    pass
+    def __init__(self, message, line=None, col=None, err_type=None, expected_obj_cls=None):
```
igordejanovic: There are PEP-8 violation at multiple places. Try to enable linter in your editor.
If you have trouble let me know and I'll fix it afterwards.

textx/model.py
```diff
@@ -509,5 +555,15 @@ def call_obj_processors(model_obj):
         if parser.debug:
             parser.dprint("CALLING OBJECT PROCESSORS")
         call_obj_processors(model)
+    
+    if type(model) not in PRIMITIVE_PYTHON_TYPES:
+        # Cross-references for go-to definition language server support
+        # Already sorted based on ref_pos_start attr (required for binary search)
+        model._pos_crossref_list = pos_crossref_list
```
igordejanovic: I think we should control this mechanism using parameter during meta-model creation. The default would be to turn it off. The rationale is that this might consume a lot of memory for larger models while it will be usable just for textX tools at the moment.

igordejanovic (changes requested):
Overall looks good but please see the comments.

danixeee:
- Fixed pull request issues
- Added functional test

igordejanovic (approved):
Looks good. Merging. Thanks.

PR #60, opened 2018-03-22
Scoping, multi file resolution bugfix

goto40:
I realized some issues with multi file reference resolution.

I like to discuss the use of one separate parser per model file (independent from the fact if the models are from the same meta model).

Changes and rationale: see CHANGELOG.md.


igordejanovic:
@goto40 Hi Pierre. Just to let you know that I don't have much time at my disposal at the moment but I'll review this PR as soon as I can. Thanks for the contributions and sorry for my slow responses.

goto40:
Hi Igor,

No problem. Take your time. No hurry! Let's wait until there is time for
proper reviews. I prefer quality and common understanding to quick shots.

Best, Pierre

Am 27.03.2018 22:14 schrieb "Igor Dejanović" <notifications@github.com>:

goto40:
I also added the ability to render a model (export.model_export) which is distributed across different files. An example/test is also provided (test_scoping/test_model_export). See last commit comment...

tests/functional/test_scoping/test_children.py
```diff
@@ -32,7 +32,10 @@
     '''
```
igordejanovic, line 32: I've noticed now that `RefClass` rule is not used anywhere. There is another usage in file `test_fully_qualified_name.py`

tests/functional/test_scoping/test_exception_from_included_model.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import unicode_literals
+from textx import metamodel_from_file, get_children_of_type
```
igordejanovic: There are several PEP-8 issues in this file. For example, `get_children_of_type` is not used.

tests/functional/test_scoping/metamodel_provider3/Base.tx
```diff
@@ -0,0 +1,11 @@
+Cls: "class" name=ID (
+( "extends" extends+=[Cls][',']+)?
```
igordejanovic: `+` at the end of `extends+=[Cls][',']+` is superfluous as it is implied by `+=`

igordejanovic: Rule `Cls` might be better indented for readability.

tests/functional/test_scoping/test_exception_from_included_model.py
igordejanovic: There are some PEP-8 issues in other files too. I can fix those after the merge.

igordejanovic (changes requested):
Excellent job Pierre!
There only few small fixes. Please see inline comments.

igordejanovic:
@goto40 BTW, I think that this multi-(meta)model support deserves a mention in `Feature highlights` in `index.md` :)

igordejanovic:
@goto40 and also, multi-(meta)-model support is such an important addition to textX that it would be awesome to have a dedicated section in the docs with examples (probably some examples used in tests) and explanations.

goto40:
I will do something we can discuss... Best, Pierre

Igor Dejanović <notifications@github.com> schrieb am Mo., 16. Apr. 2018,
12:11:

tests/functional/test_scoping/test_children.py
goto40: removed dead code in grammar (2x)

tests/functional/test_scoping/test_exception_from_included_model.py
goto40: I think I fixed many of these...

tests/functional/test_scoping/metamodel_provider3/Base.tx
goto40: Ok. Fixed.

goto40:
I worked in the in-line comments (done; maybe check for the PEP-stuff).

I added a new test (I understood something while reading your comments): test_reference_to_buildin_attribute.py

I did not yet work on the docu part (in progress).

goto40:
I have one big issue/idea (for the next pull request) which influences in parts the multi meta model feature... I will - for the moment - ***ignore these ideas and write some documentation for the current state***, since I have many open points for the concrete design/solution of my wish...

# How to extend a Meta model defined in another (installed) python package?

***Use case***: a meta model installed via pip should be extended. The installed meta model could be part of a language (e.g. some script language). The goal is to define a new meta model extending the installed meta model. In this example this new meta model could be used be to define a test specification referencing functions defined in the installed/extended/used language. 

What should be inherited from the metal model to be extended?
  - the grammar: ok  
  - object processors (e.g. validators): must be re-added manually. 
  - scope providers (same as for object processors)

Maybe it is of interest to include (parts of) a grammar for referencing purposes only. This means to inhibit creating an instance of that grammar rule in the extended meta model and, thus, only allowing to create references for that inherited rule. The latter case then makes it unnecessary to inherit most of the scope providers stuff. 


## Grammar extension

Unclear how to find the grammar file in case of grammars to be imported from other python packages (e.g. import grammarname will not work out of the box). Maybe we should somehow add a get grammar method to the Meta model? However I do not see a nice way to use it? Or maybe we should add the possibility to modify the search path for the grammars? But in the latter case, it should also work with grammars hidden in a package archive (egg or whatever). 


## Object processors

No direct support for multiple processors for one grammar rule. Maybe we should allow lists of object processors in addition to object processors?


## Scope providers

When redefining rules (creating a rule with the same name as an inherited one), we could allow in the context of the scope provider mapping to allow to use the full rule names (Grammarname.rulename) to identify both rules separately (both: the rule in the grammar to be extended and the redefined rule). However, I think this case is quite exotic and can avoided easily.

goto40:
If the last comment is confusing, ignore it for the moment, please (I just wanted to write down this ideas). Maybe it takes longer to find a proper solution... Or is it no problem at all.

I will just write some docu for the current state...

igordejanovic:
@goto40 I was thinking about similar ideas a lot in the past. How to deploy, find, inherit, extend existing languages in the system.

For discovering existing languages in the system [pkg_resources](http://peak.telecommunity.com/DevCenter/PkgResources) can be used. See [this little tool](https://github.com/textX-tools/textx-tools), that discovers existing languages and can create a scaffolding project for the given language. For example, [this language](https://github.com/igordejanovic/textx-lang-er) registers itself in the [setup.py here](https://github.com/igordejanovic/textx-lang-er/blob/master/setup.py#L58) and is discoverable by [textX tools](https://github.com/textX-tools/textx-tools). The tool finds all languages using [textX language API](https://github.com/igordejanovic/textX/blob/master/textx/langapi.py#L12). We could make extension points for whatever we would like to be discoverable, like meta-models, grammars etc.

Maybe it would be the best to make a new issue to discuss.

igordejanovic (approved):

igordejanovic:
Looks good. Merging.

PR #63, opened 2018-04-18
More docu: multi meta model support

goto40:
more docu


igordejanovic:
Looks good. Thanks! Merging.

Issue #67, opened 2018-06-04
pos_to_linecol broken in textX-1.7.0 


antoinet:
Hello Igor,

I am using `pos_to_linecol` to get line and column numbers to create human readable feedback when a parsing error occurs. Migrating from 1.6.1 to 1.7.0 I encounter the following error, using the sample code below:

```python
import textx.model
from textx import metamodel_from_str


mm = metamodel_from_str("""
Model: shapes*=Line;
Line: 'line' start=Point '/' end=Point;
Point: x=INT ',' y=INT;
""")
model = mm.model_from_str("line 1,1 / 2,2")

for shape in model.shapes:
    print(mm.parser.pos_to_linecol(shape._tx_position))
```

And the error is:
```
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    print(mm.parser.pos_to_linecol(shape._tx_position))
  File "/private/tmp/venv/lib/python3.6/site-packages/arpeggio/__init__.py", line 1612, in pos_to_linecol
    if not self.line_ends:
AttributeError: 'TextXModelParser' object has no attribute 'line_ends'
```
Seems to be a problem in arpeggio?

igordejanovic:
Hi Antoine,

thanks for the report. This seems to be a problem introduced with multi-model support. Meta-model parsers are now cloned to parse each file/string, so caching of `line_ends` used to speed up later finding of object position broke as the input is not parsed with the `mm.parser`. I think that the best solution is to use parser reference from the model instead like this:

```python
for shape in model.shapes:
    print(model._tx_parser.pos_to_linecol(shape._tx_position))
```
as the parser kept on meta-model is just a prototype parser used for cloning.

Maybe @goto40 could take a look.

Anyway, this is at least a bug in the docs so we'll keep this open until fixed.

antoinet:
Thanks, it works like you described.

goto40:
I will have a look

igordejanovic:
Closed via #68

PR #68, opened 2018-06-05
Issue #67

goto40:
* renamed metamodel.parser to metamode._parser
 * adapted code
 * added docu to clarify


goto40:
Sorry. I intended to make two pull requests. However, I messed up the branches while commiting...

The changes after the first commit concern another topic: some scope providers did not check the type of the referenced element. I added tests and fixed the code.

goto40:
Note on last comment: maybe an additional type check should be performed as a fallback in model.py.

igordejanovic:
Looks good. Thanks.

PR #69, opened 2018-06-07
Proposal for issue 66

goto40:
I added the ability to use search directories.
However, in this case I disabled globbing (see docu: scoping.md).

PR #71, opened 2018-06-12
Regular expression with group support

goto40:
I had the idea that it would be nice to be able to formulate rules with regular expressions containing one group, e.g. to parse a multiline string, with only the content of the string being inserted in the model.

I am aware that this can also be achieved with a proper post processing. I would like to discuss if this is a good idea or not.

Note: The tests only pass with some minor Arpeggio adaptation (including the matcher object in the terminal): See https://github.com/goto40/textX/tree/master-regexp-with-goto40-arpeggio for a modified travis configuration...

If this idea seems useful to you (it is okay to state that object_processors will do the job), then I can also create a pull request for the Arpeggio project.


igordejanovic:
@goto40 Just to let you know that I'm on a vacation. I'll be back by the end of the month and will take a look at this and #69. :)

textx/metamodel.py
```diff
@@ -101,12 +101,14 @@ class TextXMetaModel(DebugPrinter):
             should interact), this attribute must be set via an optional
             constructor parameter "global_repository=True" or
             "global_repository=GlobalModelRepository()".
+        replace_regexp_with_groups (bool): if True, regexp terminals are
```
igordejanovic: Since this is in action when only one group is used I think singular name would be better `replace_regexp_with_group`.
Or just `use_regexp_group`.

tmp.py
```diff
@@ -0,0 +1,74 @@
+from textx import metamodel_from_str
```
igordejanovic: This file is here by mistake I suppose.

But, it seems it would be a very nice addition to the examples. :)

docs/grammar.md
```diff
@@ -120,6 +120,28 @@ There are two types of match expressions:
...
+
+  An example model could be
+
+        data = """
+        This is a multiline
+        text!
+        """
+
+  When creating a metamodel with this grammar and the option
+  replace_regexp_with_groups enabled, a multlline string delimited with '"""'
```
igordejanovic: Typo `multlline` -> `multiline`

igordejanovic (changes requested):
Looks good overall. Some minor changes. 
Please send PR to the Arpeggio project.

textx/metamodel.py
goto40: OK.

tmp.py
goto40: Removed file.
This will be for a later pull request (I would like to add the ability to add lists of object_processors) - then we can have multiple validators (logical rules) for one grammar rule... Details are not finished.

docs/grammar.md
goto40: done

tmp.py
goto40: for this PR: OK

Issue #72, opened 2018-06-20
Regression issue


akiannillo:
I have this snippet of code 
```python
    self.meta_model = metamodel_from_file(grammar_file, classes=[Attribute] )
    processors = \
        {'Attribute':attribute_processor,
         'Subscript':subscript_processor}
    self.meta_model.register_obj_processors(processors)
```
working perfectly in version _1.6.1_. Now, in version _1.7_, the processors are not called at all.

zilakmarian:
I am having some problems with object processors too since switching to version 1.7.
Simple object processors are being run as expected. But when using more complex object processing sometimes they never start.
In object processor I am parsing another string using same metamodel and based on result changing object being processed.
In version 1.6.1 order of function calls is:
```
// objectA is found in objectB
INIT objectA
INIT objectB
PROCESS objectA
// processing objectA: parse new objectA2 with objectC within
	INIT objectC
	INIT objectA2
	PROCESS objectC
	PROCESS objectA2
PROCESS objectB
```
And in version 1.7 order looks like this:
```
INIT objectA
INIT objectB
PROCESS objectB
```
Where INIT is `__init__` of my custom object and PROCESS is object processor for that object.
Note that no processor for objectA was called.
My grammar is fairly complex now and at this moment I am not able to provide a code snippet that would replicate the problem.
I will try to revisit this issue later and find a working example.

goto40:
@akiannillo: Could you provide a sketch of you custom "Attribute"-Python class? Do you modify the object in the processor, if yes, how?

akiannillo:
@goto40 : the class is very simple
```python
class Attribute(object):
  def __init__(self, parent, value, attr, mod, prime):
    self.parent = parent
    self.value = value
    self.attr = attr
    self.mod = mod
    self.prime = prime
```
Even using a simple processor just to print out a debug message, it does not start at all.

goto40:
Hmm... I tried to setup a test to reproduce the effect. The test seems to be too simple. If you have time I would greatly appreciate more code to get the error.

zilakmarian:
I've hacked together some ugly code that can replicate this problem. I tried this with python2.7.15 and python3.6.5.
```
from textx.metamodel import metamodel_from_str

class OuterObject(object):
    def __init__(self, parent=None, col_items=None):
        print("INIT OuterObject")
        self.col_items = col_items

    def process(self):
        print("PROCESS OuterObject")

class InnerObject(object):
[...]

PR #74, opened 2018-06-23
Possible idea for issue72

goto40:
Possible quick fix. Is it meant to call object processors for base classes? If no, the concrete instance must be used - else the base classes must be taken into account to. In the latter case I am not sure if all base classes are considered... 

This pull request must be analyzed with care to check if the solution is only a quick fix or a valuable commit...


goto40:
In https://github.com/igordejanovic/textX/blob/v1.6.1/textx/model.py.

the object processor call logic deduce the metaclass from the class name of the current object being processed.

That explains the difference. The quick fix of this pull request adds the old behavior from 1.6.1. if no match for the class based on the Metamodell is found.

igordejanovic:
@goto40 Thanks for the fix. Yes, in this case one can register obj processor for base class used in grammar reference. If processor for that base class is not found concrete class is used. You are right, other base classes down the hierarchy are not tried but I think this solution is good as it retains semantics from 1.6.1 while adding possibility to also register processors using base classes.

refs #72

PR #77, opened 2018-07-20
New multi meta model support for references-only 

goto40:
This commit is about allowing textx **meta models to be more modular** (in terms of **deploying a meta model as python library**, hiding all grammar, scoping, validation, user classes etc).

A new **meta model can now reference elements from another meta model** (e.g., from a lib) without having to care about how exactly the original meta model is constructed.

The change is about **a few lines only (lang.py and metamodel.py)**, to correctly handle the classes of referenced elements in lang.py.

The changes in detail:
 * docu changes (multimetamodel.md)
 * test cases (especially tests/functional/test_metamodel/test_multi_metamodel_refs.py)
 * **lang.py**: correctly identify meta classes fro references
 * **metamodel.py**: 
    * manage referenced meta models in a new list
    * help lang.py to find the correct rule/meta class
 * scoping./init.py: support for allowing to add individual models for lookup (instead of only loading models from files; useful for model_to_str)
 * scoping/providers.py: 
   * bugfix in line 259, not identified until now by unittests
   * allow adding individual models for lookup
 * misc: some PEP fixes


igordejanovic:
I'm thinking could we make a more flexible extension mechanism where one could extend the meta-model without access to the constructing elements (grammar, processors etc.)? Furthermore, could we integrate both use-cases, i.e. extending and referencing?

Let's say that we use `pkg_resources` extension points to register and discover textX meta-models at run-time, like it is done in [textx-tools](https://github.com/textX-tools/textx-tools) (see how [extension point is registered in some of the language projects](https://github.com/igordejanovic/textx-lang-er/blob/master/setup.py#L58)). It is enough for a library to be installed via pip for its meta-models to be discoverable by a project like it is done [here](https://github.com/textX-tools/textx-tools/blob/master/txtools/gen.py#L12).

Now, we could reuse keyword `import` but extend its syntax and semantics to allow for importing of both local grammars and installed meta-models (e.g. by prefixing FQN with `meta:` for example). Imported meta-model elements could both be referenced from our new meta-model and overridden if the same rule name is used (maybe we could introduce some annotation on the rules to make that explicit - like `overrides`).

A classic problem with multiple inheritance arise in case of multiple imports, but we could use similar approach to Python MRO to search for a proper meta-class given its name. The same can be used for specification of object processors, user classes etc. If we supply a new user class we override the user class from the target library.

A subtle difference between extending and referencing is the notion of the root rule. In the case of extension the root rule remains the same while in referencing the root rule is the first rule from our grammar. Another difference that one might argue about is the question "What happens to the target meta-model reference resolving in both cases?". In the case of extension all target meta-model references should use overridden meta-classes and fall back to their own if there is no overridden one, but in the case of referencing we might opt for a target meta-classes to always resolve using target meta-model. Could we say that the first behaviour is always preferred? I don't know.

The mayor obstacle I see is manipulating underlying Arpeggio parser as we must come up with a single parser based on our resolution. Luckily, since Arpeggio is PEG and each parser rule is a node in the parser model this shouldn't be a big problem but would require changes in Arpeggio to allow for such manipulation.

I hope that this brain-dump makes some sense :)

goto40:
I fully agree, that extending a meta model is much more complicated than just to allow a new meta model to reference an existing one. My intention here, is to account for the need of such references. 

A **possible use case** is the following: a data format DSL (D) defines messages, while a data flow DSL (F) allows to define data flows. F in this case is using data elements from D. There is no need for F to know any details of D, except the fact that data elements exist. There is a **minimal dependency** between both meta models: separate scoping, validation, etc. 

The **current ability of TextX to include other grammars** allows pretty well to implement such an example sketched above. However, the language designer has much more possibilities, which would increase the depency (e.g., allow to define a message in the data flow language itself). The **new feature proposed** here, allows the desired references  to the data model but prohibits the (much more difficult to master) extension and reuse of the data model. 

For the **discovery** of languages: I see big advantages for tools, like editors, to be able to discover meta models, generators, etc. A language designer, however, exactly knows what meta model should be referenced. The additional dependency to some discovery mechanism (with potential name clashes) does not yield an adequate advantage (in my eyes, I surely do not know about many use cases out there). But the **proposed pull request does not conflict with a future grammar based meta model inclusion** for references: meta models discovered from within the grammar specification could be included the same way as it is proposed here. 

I propose to further discuss the pull request as it is. We should definitely pursue the ideas concerning extension aspects you mentioned in a separate thread...

tests/functional/test_metamodel/test_multi_metamodel_refs.py
```diff
@@ -0,0 +1,116 @@
...
+grammarBWithImport = """
+Model: imports+=Import b+=B;
+B:'B' name=ID '->' a=[A];
+Import: 'import' importURI=STRING;
+"""
+
+
+def test_multi_metamodel_references1():
+    global_repo = scoping.GlobalModelRepository()
+    mm_A = metamodel_from_str(grammarA, global_repository=global_repo)
```
igordejanovic, line 28: @goto40  Does this means that in order to be *referencable* meta-model needs to be instantiated with a shared global repository? I wonder if one is going to reference meta-model from a third-party library how that is going to be achieved? 

I'm trying to understand the use case where one language developer is distributing his languages as a Python library and the another developer wants to reference elements contained in a model parsed by third-party meta-model/parser.

textx/scoping/providers.py
```diff
@@ -340,6 +341,17 @@ def _load_referenced_models(self, model):
         for filename_pattern in self.filename_pattern_list:
             model._tx_model_repository.load_models_using_filepattern(
                 filename_pattern, model=model, glob_args=self.glob_args)
+        for m in self.models_to_be_added_directly:
+            model._tx_model_repository._add_model(m)
+
+    def add_model(self, model):
+        """
+        Adds a model directly. Useful when combining models
+        pasred from a string (instead of a file).
```
igordejanovic: Typo: `pasred` -> `parsed`

igordejanovic:
> I propose to further discuss the pull request as it is. We should definitely pursue the ideas concerning extension aspects you mentioned in a separate thread...

I agree. PR looks good. I have a question. Please see above.

tests/functional/test_metamodel/test_multi_metamodel_refs.py
goto40: An example is added to unittests, where three libs are simulated and the second lib uses the first and the third lib uses the second.

It also illustrates how to achieve a single global repo shared with an appropriate third-party lib.

However, the example illustrates, that a global_repo is not required in cases, where one model (including many others) is loaded... (comparison included in unittest). I also changed the docu (the global repo text was not 100% correct - and confusing).

textx/scoping/providers.py
goto40: some types fixed.

goto40: typos...

goto40:
I will also add some validation code to the meta models...

goto40:
I had to modify model.py to correctly call the obj_processors in case of multi model structures. This should have no influence on the performance of non-multi model structures...

tests/functional/test_metamodel/test_multi_metamodel_refs.py
goto40: To make things clear:
 * Does this means that in order to be referencable meta-model needs to be instantiated with a shared global repository?
   * No.
 * I wonder if one is going to reference meta-model from a third-party library how that is going to be achieved (*to have a common repo*)?
   *  This only works, if the third party meta model already has a global repo. Then it can be achieved by passing the global_repo of the referencable meta-model to the new meta-model (like in the new unittest example).
   * But, as the unittest demonstrates, this is not required

goto40:
In general, it seems to be unclear to a user, when a global repo makes sense... This should be explained in a separate thread in detail (+more docu/examples). To sum up: global repos are only required, when different model_from_file-calls should share common included files.

tests/functional/test_metamodel/test_multi_metamodel_refs.py
igordejanovic: OK. That makes sense.

tests/functional/test_metamodel/test_multi_metamodel_refs.py
```diff
@@ -0,0 +1,346 @@
...
+                              global_repository=global_repo,
+                              referenced_metamodels=[mm_A])
+
+    # define a default scope provider supporting the importURI feature
+    mm_B.register_scope_providers({"*.*": scoping_providers.FQNImportURI()})
+
+    # map file endings to the meta models
+    scoping.MetaModelProvider.clear()
+    scoping.MetaModelProvider.add_metamodel("*.a", mm_A)
+    scoping.MetaModelProvider.add_metamodel("*.b", mm_B)
```
igordejanovic, line 104: Could we make this API a bit simpler for the user? If functions of `MetaModelProvider` are static we could add additional convenience function `add_metamodel` in the `scoping` module. So user could call `scoping.add_metamodel` in addition to a more verbose way as it is now. Also, we could add `scoping.clear_metamodels`. I think this could make it a bit easier to use and to remember.

textx/scoping/providers.py
```diff
@@ -348,7 +360,7 @@ def load_models_in_model_repo(self, global_model_repo=None):
         Normally this is done automatically while
         reference resolution of one loaded model.
 
-        However, if you wich to load all models
+        However, if you which to load all models
```
igordejanovic: which -> wish? :)

igordejanovic:
@goto40 Looks good. Great work! I have a suggestion about API. Please see above. That can be a part of another PR. Feel free to merge this when you are ready.

igordejanovic (approved):

tests/functional/test_metamodel/test_multi_metamodel_refs.py
goto40: Hmm... Yes, I postpone this API change. 

The static methods are definitely a weak point (global variable) and further establish them by adding the proposed methods will not make it better.

I will maybe think about if I find a better (more local) solution without littering the global namespace...

textx/scoping/providers.py
goto40: Jepp. Fixed.

goto40:
I merge now.

PR #79, opened 2018-07-23
Master issue78

goto40:
This PL represents
 * a possible solution to issue 78 (see first test case in test_issue78.py which worked for textx v1.6.1)
 * a possible extension to be able to (correctly?) handle object processors in case of ABSTRACT_RULEs, when processors are define for the "Base class" and the "Specialized classes".

Before going into the changes, I want to state that I think there should be no performance impact compared to the previous version (when no such multiple object processors are defined): there is no such thing as a new for-loop or similar things. Of course more processors are called now, since (obviously) issue 78 shows that some processors are omitted compared to v1.6.1.

The changes (all in model.py, call_obj_processors)
 * **issue78** (test_issue78_obj_processors_base_attr_proc()): in model.py, call_obj_processors **now uses the metaclass of the model_obj under analysis**, instead of the metaclass derived from the grammar (which could be a "base class" in case ABSTARCT_RULE). This is required, since else attributes from the "specialized class" will not be processed.
 * Extra addition (test_issue78_obj_processors_order_of_eval()): when **multiple object processors can be identified, the processor for the "specialized class" is called first**, followed by the processor of the "base class".
 * Extra addition (test_issue78_obj_processors_replacement*()): when **multiple object processors return a value, the value of the "specialized class"'s processor dominates** (although both processors are called).

All existing projects should have no impacts, when no such processors for "base" and "specialized classes" are defined at the same time...

Note (minor): it is not 100% clear to me, why the MetaAttr-flag "ref" is True for "normal" attributes to be processed by call_obj_processors (this aspect of the code is unchanged). Maybe a glossary defining a "containment", a "reference" and a "containment reference" would help me out...

### DIFF
--- a/tests/functional/regressions/test_issue78.py
+++ b/tests/functional/regressions/test_issue78.py
@@ -1,0 +1,21 @@
+import textx
+def test_issue78_obj_processors_base_attr_proc():
+    mm = textx.metamodel_from_str('''
+        Model: base+=Base;
+        Special1: '1' name=ID;
+        Special2: '2' name=ID test=TestObj;
+        Base: Special1|Special2;
+        TestObj: name=ID;
+    ''')
+    test_list=[]
+    mm.register_obj_processors({
+        'TestObj': lambda o: test_list.append(o.name),
+    })
+    m = mm.model_from_str('''
+    1 S1a
+    1 S1b
+    2 S2a t2a
+    2 S2b t2b
+    ''')
+    assert ['t2a','t2b'] == test_list
+
--- a/tests/functional/regressions/test_issue78.py
+++ b/tests/functional/regressions/test_issue78.py
@@ -147,10 +147,9 @@ def test_issue78_quickcheck_no_obj_processors_called_for_references():
     assert ['a1','a2','a3'] == test_list
 
     # only references to A: --> obj proc not called
-    test_list=[]
     global_repo_provider.add_model(m1)
     m2 = mm.model_from_str('''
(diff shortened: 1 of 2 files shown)


igordejanovic:
> Note (minor): it is not 100% clear to me, why the MetaAttr-flag "ref" is True for "normal" attributes to be processed by call_obj_processors (this aspect of the code is unchanged). Maybe a glossary defining a "containment", a "reference" and a "containment reference" would help me out...

I definitely agree that we should have some dev docs. I was just trying to recover the semantics of those three different kind of relationships and it took me a while. 

So here it goes:

screenshot_20180724_102853

Case (1) is a containment, case (2) is a containment reference, case (3) is a reference.

In case (1) object A owns object B and object B exists "inside" object A. You can think of this as e.g. attributes of primitive types in Java. In case (2) we have 2 "real" objects but object A still owns object B. When object A dies so should B. Case (3) is just an ordinary reference without any life-cycle connection between A and B. B can exists on its own and doesn't die when A dies.

Now, when I look at it I think that `if` in `call_obj_processor` could be reduced from:
```python
            if metaattr.is_base_type or (metaattr.ref and metaattr.cont):
```
to 
```python
            if metaattr.cont:
```
So the point of this is to drill down only following the containment reference to prevent visiting objects more than once.

igordejanovic:
@goto40 Looks good. I agree with the proposed changes. Please feel free to merge it.

goto40:
I merge the request. Thank you for the detailed infos.

PR #84, opened 2018-08-29
Contributing and pep8 fixes

igordejanovic:
@goto40 This PR fixes numerous PEP-8 violations in the tests and brings a new contribution guide in the docs. This should help new contributors get started easier. Please review when you find time.

I'm planing to make flake8 adherence mandatory trough Travis CI.


tests/functional/test_scoping/test_import_module.py
```diff
@@ -115,16 +115,16 @@ def test_model_with_imports_and_errors():
...
             "/interface_model1/model_b/app_error1.if")
 
     with raises(IOError, match=r'.*file_not_found\.if.*'):
-        _ = my_meta_model.model_from_file(
+        my_meta_model.model_from_file(
             abspath(dirname(__file__)) +
             "/interface_model1/model_b/app_error2.if")
 
-    #################################
+    #####tests/functional/test_scoping/test_children.py############################
```
goto40: Type/copy paste effect.

igordejanovic: Ah, yes, good catch :)

goto40:
Very good! I will take my time to read and test your proposal.

At first sight: Do we also need to include the flake command in the .travis.yml file?

igordejanovic:
> At first sight: Do we also need to include the flake command in the .travis.yml file?

Yes, you are right, that is something I was up to do next. I'll need to investigate what is the best way to do it.

goto40:
I tested the contribution manual, and no errors occurred.

goto40:
Maybe we could also start to collect architectural entry points... Such as
 * structure: overall file/module structure (with rough overview about responsibilities)
 * activity: parsing and model building / scope resolution
 * ...

Do you think something like doxygen could help to link architectural snippets to the code?

You remember the explanation you gave me concerning 'containment' and 'reference' in PR #79. This could be valuable to other contributors.

igordejanovic:
Thanks for testing. Merging.

Regarding the dev docs, I agree. The overall architecture/processes description and main internal APIs would be helpful, not only for new contributors but for ourself as well as our brains are not the best places to keep all those stuff :)  Although, I wouldn't go into too much of  details (just the high-level overview) as the docs (especially dev docs) tend to become out-of-sync very fast.

OTOH, I'm not that eager to introduce new documenting tools. The project is using mkdocs which is working really good and is very simple to use so IMHO we should use it for all docs parts.

Issue #89, opened 2018-10-15
Breaking API changes


ghost:
It seems that in version 1.7.0 there are changes breaking the compatibility with the previous version (certtools/intelmq#1246).

The specific error that we get is:
> AttributeError: module 'textx.model' has no attribute 'metamodel'

Maybe `textx.model.metamodel()` is now `textx.model.get_metamodel()`

And then
> AttributeError: 'TextXModelParser' object has no attribute 'line_ends'

(and probably more)

Is there a stable API which can be used? Is there a list of removed methods with their corresponding replacements?

igordejanovic:
Hi @wagner-certat. Thanks for the report.

We are trying to keep API as much stable as possible but backward compatibilities creep in occasionally which should be noted in changelog. If not, that is a bug. Version 1.7.0 was a release with a lot of changes so we might miss to list some in the changelog. BTW, there is a bugfix release 1.7.1 so I suggest you to go for that version although I don't think that would solve this issue.

We decided recently to apply semantic versioning and increase major version number at each breaking change so that should better communicate changes severity.

Could you provide more info about the error your are experiencing? At least exact Arpeggio version and the full stack trace of the error. Also if you could outline the general usage pattern, that would also be helpful in tracing down the problem.

igordejanovic:
There are breaking API changes from issue #48 that seems not registered in the changelog. Will fix it. That is the first part of the problem. But for the second would need more input.

ghost:
> BTW, there is a bugfix release 1.7.1 so I suggest you to go for that version although I don't think that would solve this issue.

I re-tested it with the current 1.8.0 release, but it was also a problem with 1.7.1.

> We decided recently to apply semantic versioning and increase major version number at each breaking change so that should better communicate changes severity.

Great to hear :)

> Could you provide more info about the error your are experiencing? At least exact Arpeggio version and the full stack trace of the error. Also if you could outline the general usage pattern, that would also be helpful in tracing down the problem.

The Arpeggio version is the recent one, 1.9.0.

The (relevant part of the) traceback from the linked issue:
```
  File "/home/travis/build/certtools/intelmq/intelmq/bots/experts/sieve/expert.py", line 323, in get_linecol
    metamodel = textx.model.metamodel(model_obj)
AttributeError: module 'textx.model' has no attribute 'metamodel'
```
With this code we want to get the position of a rule in the model.

[...]

PR #92, opened 2018-10-15
flake8, coverage, etc. and minor code updates

goto40:
To push further the quality activities, I continued - as you suggested - the usage **flake8** in travis: 
 * I fixed many  minor aspects (empty lines etc.)
 * I excluded six.py and lang.py (these files must be analyzed separately)
 * I had two possibly problematic cases:
   * examples/StateMachine/state_machine.py: "except" with Exception (should be okay?) 
   * examples/StateMachine/smdot.py: unclear if "\\" is correct (in doubt I remove my change)

Inspired by #89 (I saw they have a coverage output) - added **coverage** to our dev-dependencies and included the command in travis.yml: You can also interactively inspect the coverage with "coverage html"...

What do you think about it?


igordejanovic:
@goto40 Thanks! I gave it a quick look and it looks really good. Will test a bit more later today.

igordejanovic:
@goto40 Looks great. Feel free to merge.
I noticed that you got idea project file committed accidentally. I suggest to add `.idea` to the `.gitignore` to prevent that from happening.

PR #94, opened 2018-10-17
Some howto snippets...

goto40:
Hi - 
If you have time...

 * How do you think we should incorporate this? See mkdocs.yml
 * Do we need more prosa? or introductory text?
 * Typos???


goto40:
flake8 etc. already striked back :-)

igordejanovic:
Thanks. This looks good. Really like this to-the-point examples. I don't think we should put in more introductory text. I like the short form :)
Maybe just change the title in the .md file  to be `HowTos`.

> flake8 etc. already striked back :-) 

Good to know it is working :)

Merge when you feel it's ready and I'll regenerate the docs.

goto40:
Ok. I merge now

PR #98, opened 2018-11-08
Issue97 + Issue95

goto40:
A possible solution to issue 97. To be discussed:
 * The problem is that a FLOAT needs to match an INT as well (to be backward compatible and to fulfill "intuitive modeling").
 * If we wish not to add some hacks while creating the model from the parsetree, we cannot distinguish a INT from a FLOAT.
 * I **suggest in this pull request** to add a STRICTFLOAT as a specialization of NUMBER (instead of FLOAT), which make the classes distinguishable. Since the basetypes are always transformed to python objects, I expect no further effects for users.
 * The only effect, is that a NUMBER can return an INT now (this I adjusted 1-2 testcases)  

Further observations:
 * Python 3.3 seems to make problems? See travis log
 * I missed the source of images/base_types.png (needs to be adapted, if the solution is adequate)


goto40:
Note: the master also fails for Python 3.3 (I restarted the travis build).

igordejanovic:
I agree with the solution. Feel free to merge when ready.

> Python 3.3 seems to make problems? See travis log

setuptools seems to require 3.4 as minimum version now. It is time for us to drop 3.3 support so I did it on the master. It is still failing due to flake8 error but I see that you resolved that on this branch already so if will be fixed when this is merged.

> I missed the source of images/base_types.png (needs to be adapted, if the solution is adequate)

I think I just cut it out from a model produced by `textx visualize`. Not sure what to do with this as `FLOAT` is not anymore a real subtype of `NUMBER`. It should be conceptually but only the match is different. `STRICTFLOAT` is a special kind of `FLOAT` although this link is not really an inheritance. I guess we could leave the base type image as it is, maybe just make a note near the image that actually `STRICTFLOAT` is used for matching of `NUMBER`. Or we could modify the title of the `FLOAT` class in this image to be `(STRICT)FLOAT`.

PR #106, opened 2018-11-14
Issue103 import uri ext

goto40:
Allow python like imports:
```
import folder.module as m
ref m.obj
```

In particular, this extension allows to
 * transform the importURI attribute "a.b" --> e.g., "a/b.ext".
 * allow to control the lookup rules (m.obj instead of obj in the example above).

Affected textx files:  textx/scoping/*

Adapted documentation (scoping.md)

Added new unittests.

### DIFF
--- a/tests/functional/regressions/test_issue103_python_like_import.py
+++ b/tests/functional/regressions/test_issue103_python_like_import.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
 from textx import metamodel_from_str
-from os.path import dirname, abspath
+from os.path import dirname, abspath, join
 import textx.scoping.providers as scoping_providers
 
 
@@ -39,7 +39,7 @@ def test_issue103_python_like_import():
     #################################
 
     m = mm.model_from_file(
-        abspath(dirname(__file__)) + "/issue103/main.mod")
+        join(abspath(dirname(__file__)), "issue103", "main.mod"))
 
     #################################
     # TEST MODEL
--- a/tests/functional/regressions/test_issue103_python_like_import.py
+++ b/tests/functional/regressions/test_issue103_python_like_import.py
@@ -2,7 +2,6 @@ from __future__ import unicode_literals
 
 from textx import metamodel_from_str
 from os.path import dirname, abspath
-import textx.exceptions
 import textx.scoping.providers as scoping_providers
 
 
@@ -17,11 +16,8 @@ def test_issue103_python_like_import():
                 classes*=Class
                 vars*=Var
         ;
-        
         Class: 'class' name=ID '{' '}' ';';
-        
-        Var: 'var' name=ID '=' 'new' class=[Class|FQN] '(' ')';
-        
+        Var: 'var' name=ID '=' 'new' theclass=[Class|FQN] '(' ')';
         FQN: ID+['.'];
         Import: 'import' importURI=STRING;
         Comment: /#.*$/;
@@ -34,7 +30,8 @@ def test_issue103_python_like_import():
         return m.group(1)
 
     mm.register_scope_providers(
-        {"*.*": scoping_providers.FQNImportURI(importAs=True, name_setter=name_setter)})
+        {"*.*": scoping_providers.FQNImportURI(importAs=True,
+                                               name_setter=name_setter)})
 
     #################################
     # MODEL PARSING
@@ -47,7 +44,7 @@ def test_issue103_python_like_import():
     # TEST MODEL
     #################################
 
-    #assert my_model.packages[0].name == "B"
+    assert m.vars[0].theclass.name == "a"
 
     #################################
     # END
--- a/tests/functional/regressions/test_issue103_python_like_import.py
+++ b/tests/functional/regressions/test_issue103_python_like_import.py
@@ -24,10 +24,10 @@ def test_issue103_python_like_import():
         ''')
 
     def name_setter(import_obj):
-        import re
-        m = re.search(r'^([^\.]+).*$', import_obj.importURI)
-        assert m
-        return m.group(1)
+        # this method is responsible to deduce the module name in the
+        # language from the importURI string
+        # e.g. here: import "file.ext" --> module name "file".
+        return import_obj.importURI.split('.')[0]
 
     mm.register_scope_providers(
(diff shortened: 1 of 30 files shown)


goto40:
The last two commits (automatic naming of importURI objects, required for the issue103-example) has to be discussed: The name_setter is not the nicest solution. But works (and is somehow similar to the importURI_converter logic)...

tests/functional/regressions/test_issue103_python_like_import.py
```diff
@@ -0,0 +1,54 @@
...
+        Var: 'var' name=ID '=' 'new' class=[Class|FQN] '(' ')';
+        
+        FQN: ID+['.'];
+        Import: 'import' importURI=STRING;
+        Comment: /#.*$/;
+        ''')
+
+    def name_setter(import_obj):
+        import re
+        m = re.search(r'^([^\.]+).*$', import_obj.importURI)
```
igordejanovic: If I understand this regexp, you are collecting everything before the `.`. In that case much simpler would be:
```python
return import_obj.importURI.split('.')[0]
```

tests/functional/regressions/test_issue103_python_like_import.py
```diff
@@ -0,0 +1,51 @@
...
+        assert m
+        return m.group(1)
+
+    mm.register_scope_providers(
+        {"*.*": scoping_providers.FQNImportURI(importAs=True,
+                                               name_setter=name_setter)})
```
igordejanovic: I'm not happy with the name of this param. `name_setter` is too generic term. Maybe `uri_to_scope_name`? Something more specific but still short enough to be easy to type.

igordejanovic (changes requested):
Great work Pierre. As usual :)
See some inline comments.

I'm thinking about some redesign of textX language to support reference resolving which would hopefully make scoping/resolving more straightforward. Will post as a new issue for discussion.

tests/functional/regressions/test_issue103_python_like_import.py
goto40: Good point. I also added a comment to make clear what happens here...

tests/functional/regressions/test_issue103_python_like_import.py
goto40: I renamed it to importURI_to_scope_name. Good idea, else in two weeks we wonder what name_setter means... ;-)

tests/functional/regressions/test_issue103_python_like_import.py
```diff
@@ -0,0 +1,52 @@
...
+            FQNImportURI(importAs=True,
+                         importURI_to_scope_name=importURI_to_scope_name)})
+
+    #################################
+    # MODEL PARSING
+    #################################
+
+    m = mm.model_from_file(
+        abspath(dirname(__file__)) + "/issue103/main.mod")
+
```
igordejanovic, line 43: All path construction should be done by `os.path.join` to prevent cross-platform issues reported in #110. Please correct all instances for this PR and see #110 when find some time.
```python
m = mm.model_from_file(
        join(abspath(dirname(__file__)), "issue103", "main.mod")
```
where `join` is `from os.path import join`

igordejanovic (changes requested):
Please see another comment. Sorry for not reporting in the first review.

tests/functional/regressions/test_issue103_python_like_import.py
goto40: No problem ;-) I really appreciate this collaboration!

goto40: Uff. There were quite a lot of these... At least I learned something! ;-)

I am not sure if I catched all of them. I also cannot test it using windows. @igordejanovic  can you test it? Else we can ask @dkrikun (see #110).

dkrikun:
I could test it, but I'm currently a bit struggling with dev. environment
in general, see also #110.

How should one run tests btw, is it `pytest` or `setup.py test`?

goto40:
It is "py.test tests/functional".
 * You can also have a look at https://github.com/igordejanovic/textX/blob/master/docs/about/contributing.md
 * You can also have a look at ".travis.yml". There you can get hints how the system is setup

tests/functional/regressions/test_issue103_python_like_import.py
igordejanovic: @goto40 I don't do windows either ;)

goto40:
@dkrikun : if your problems concerning
```
        inner_model = my_model._tx_model_repository.all_models.filename_to_model[
>           abspath(dirname(__file__)) + "/interface_model2/model_b/base.if"]
E       KeyError: 'C:\\proj\\textX\\tests\\functional\\test_scoping/interface_model2/model_b/base.if'
```
are not getting better at all (with the new version from this pull request), tell me...

I expect some of them to go away (or even all).

Note: You may clone "https://github.com/goto40/textX.git"... and checkout the correct branch
```
git clone https://github.com/goto40/textX.git
cd textX
git checkout origin/issue103_importUriExt 
export PYTHONPATH=.
python3 -m pytest tests/functional
```
(Under Windows you may use "set PYTHONPATH=."??)

dkrikun:
@goto40 much better, only one fail, output attached
[pytest_out.txt](https://github.com/igordejanovic/textX/files/2592083/pytest_out.txt)

igordejanovic:
@goto40 This could be fixed by changing:
```
assert "inheritance/model_a.a" in text
```
with:
```
assert "inheritance{}model_a.a".format(os.path.sep) in text
```

goto40:
Thanks @igordejanovic, fixed.
@dkrikun : You can try again...

dkrikun:
@goto40 all green

igordejanovic:
If everything is ok now, @goto40 before merging could you please tidy up history a bit? You can squash last several commits related to cross-platform compatibility fixes in one commit with a proper log message and the rest of the commits, related to import feature, in another.

goto40:
I will do it... Best wishes, Pierre

Am Sa., 17. Nov. 2018, 19:45 hat Igor Dejanović <notifications@github.com>
geschrieben:

goto40:
closed in favor for #114 (squashed commits)

Issue #107, opened 2018-11-15
Issue with ordered choice and references (FQN scoping)


mfrigerio17:
Hi, consider the code:

```
import textx

def issue():
    grammar = '''
    Model:
        groupKind1 = GroupKind1?
        groupKind2 = GroupKind2?
        values += Value+
    ;

    LiteralKind1 : name=ID;
    GroupKind1:
        "Kind1" name=ID "{"
            vars += LiteralKind1*
        "}"
    ;

    LiteralKind2 : name=ID;
    GroupKind2:
        "Kind2" name=ID "{"
            vars += LiteralKind2*
        "}"
    ;

    Value : Ref1 | Ref2  ;

    Ref1 : val = [LiteralKind1|FQN] ;
    Ref2 : val = [LiteralKind2|FQN] ;

    FQN : ID'.'ID;
    '''
    mm = textx.metamodel_from_str(grammar)
    fqn = textx.scoping.providers.FQN()
    mm.register_scope_providers({"Ref1.val": fqn, "Ref2.val": fqn})

    model = '''
    Kind1 kind1 {
        a b c
    }
    Kind2 kind2 {
        a b c
    }

    kind1.a kind1.b kind2.a
    '''

    m = mm.model_from_str(model)
```

It fails with `Unknown object "kind2.a" of class "LiteralKind1`
I understand that the `Ref1` rule, that comes first, is selected because the text "kind2.a" matches it.
But why is the subsequent failure (no object found) not triggering an attempt with the other rule, `Ref2` ?

I suspect that only text-matching failures lead to the following alternatives in an ordered choice... But then how do I make my grammar work? (which obviously means recognizing that "kind2.a" is a reference to the "a" defined in the group "kind2").

Thanks in advance

goto40:
The problem in the example above, is that Ref1 and Ref2 are not distinguishable by the parser.
Thus, the first match (Ref1) is always chosen.

Note: It is not possible to select a rule by the type of a reference in the rule. The reference resolution takes place after parsing.

You could try the following (the validator is optional in my option, since the grammar does not allow
the bad cases except for the extension I added to test the validator, see token "/" in GroupKind2:

  * Ref1 and Ref2 are now merged to Ref
  * LiteralKind1 and LiteralKind2 have now a common base, referenced by Ref

```
import textx

def issue():
    grammar = '''
    Model:
        groupKind1 = GroupKind1?
        groupKind2 = GroupKind2?
        values += Ref+
    ;

[...]

Issue #102, opened 2018-11-14
Add import usage in entity example


dkrikun:
Please exercise scoping in entity example. I would love to see how one can import entity files one from another. IMO, it is something almost everybody designing a DSL will want to have.

igordejanovic:
Definitely agree on this. :)  Would be nice to have a complete example/tutorial with imports and scoping. It would make multi-(meta)-model and scoping features more visible to the users.

goto40:
Maybe this unittest could be migrated to the examples folder?
test_multi_metamodel_types_data_flow1

In

https://github.com/igordejanovic/textX/blob/master/tests/functional/test_metamodel/test_multi_metamodel_refs.py


Am Mi., 14. Nov. 2018, 10:38 hat Igor Dejanović <notifications@github.com>
geschrieben:

dkrikun:
@goto40 I'm sorry, I am new to textX and it is quite understandable and "just works", but scope providers feel like "black magic" and it is not very clear from the test code how they work and what to expect. A simple example of how to implement scoping/import mechanism would be great :)

igordejanovic:
> Maybe this unittest could be migrated to the examples folder? test_multi_metamodel_types_data_flow1 

Probably we should start with multi-model and scoping tutorial. I agree that entity example would be a good fit at it should be familiar to beginners and shows a small but complete example with code generation. I would like to take on this one but probably won't be able in a next month or so.

After that we could make a more complex tutorial/example with a multi-meta-model support. There are many nice tests that could serve as a basis for this (e.g. flow, components ...).

PR #109, opened 2018-11-16
Issue107

goto40:
Thanks for the fruitful discussion! #107 also irritated me first - until I realized there was no bug (as far as I see). However, as suggested in #102 we should investigate in tutorial to make the current feature more visible to the public - and to find more bugs!

Here, I try to make clear how to use the RelativeName Scope provider... Have a look and make suggestions what examples to add or which part pf the text to reformulate...


docs/scoping.md
```diff
@@ -115,6 +115,66 @@ We provide some standard scope providers:
...
+   or - more usual - a **list of named objects** containing the named object to 
+   be resolved.
+
+        grammar = '''
+            Model:        kinds += GroupKind values += Ref;
+            LiteralKind : name=ID;
+            GroupKind:    kindName=ID name=ID "{"
+                            vars *= LiteralKind
+                          "}";
+            Ref: ref0=[GroupKind] '.' ref1=[LiteralKind];
```
igordejanovic, line 137: This looks problematic to me as, if I understand correctly, you will end up with a list of `Ref` objects in `Model.values` and you will have to do `ref.ref1` to get to the `LiteralKind` instance you actually wanted in the `values` list. Did I understand this correctly?

goto40: This example demonstrates hat to simulate FQN lookup with step-by-step local scope provders:
 * **FQN lookup uses the natural structure of the model** (named things containing named things which maps to FQN-like mapping: name1.name2.name3...)
 * In my eyes: **local scope dependent lookup seeks to resolve one name** (not necessarily a FQN, typically only an ID) relative to the local context. 
  * In case of the Ref.ref1 the local context if Ref: it uses Ref.ref0 to find its targets - or in the case of Ref.ref0 it uses the parent of Ref to resolve the reference.
  * In https://github.com/igordejanovic/textX/blob/master/tests/functional/test_scoping/test_reference_to_nontextx_attribute.py you have an example with a custom provider (not RelativeName) - it is a 10-liner, including json-IO...: Here we map an ID to an entry in a JSON file...
 * The **RelativeName scope provider is just a shortcut** to help to implement local scope, like 
   * give me the method of the instance defined by the class, 
   * or (like in the new example) the configurations of the current scenario... 


--> **See new scenario example**
Maybe we find a better name for RelativeName?

goto40: You understood correctly: You get Refs, not ref1s...
I usually use a custom class to simplify access to to ref1 (if it is a more complicated case).

Important: the example test_issue107_example_with_relative_name is not a good one (because you should better use FQN). Maybe we should remove it...

A better example is the (new) testcase/testaspect/scenario/configuration example...

goto40: I added a comment, that you should not simulate FQN with RelativeName in the original example.

igordejanovic: Ah, gotcha. So, using relative name scope provider is misleading in this case. I think that we should provide tests/docs with best/recommended practice only as otherwise we could confuse users.

I agree that `test_issue107_example_with_relative_name` and `test_issue107_example_with_relative_name_deep_tree` should be converted to either use a FQN in a proper way or removed.
I also suggest the same for the docs example.

igordejanovic: BTW, I'm thinking in a direction of blending `FQN` and `RelativeName`. I couldn't resist the feeling that `FQN` could be just a special case of `RelativeName`. That would directly support #111

goto40: I removed the two testcases I think we do not need any more.

I copy the testcase with "parent(TYPE)". maybe I reuse it later with some modifications. We do not have any test for the "parent(TYPE)" feature yet...

```
def test_issue107_example_with_relative_name_deep_tree():
    # This example simulates the FQN scope provider with
    # the RelativeName provider...
    # It would be easier to directly use the FQN here...
    from textx.scoping.providers import RelativeName

    grammar = '''
        Model:        kinds += GroupKind values+=Formula;
        LiteralKind : name=ID;
        GroupKind:    kindName=ID name=ID "{"
                        vars *= LiteralKind
                      "}";
        Formula:      formula=FormulaPlus ';';
        FormulaPlus:  sum+=FormulaMult['+'];
        FormulaMult:  mul+=FormulaVal['*'];
        FormulaVal:   (ref=Ref)|(val=NUMBER)|('(' rec=FormulaPlus ')');
        Ref: ref0=[GroupKind] '.' ref1=[LiteralKind];
    '''
    mm = textx.metamodel_from_str(grammar)
    mm.register_scope_providers({
        "Ref.ref0": RelativeName("parent(Model).kinds"),
        "Ref.ref1": RelativeName("ref0.vars")
        })

    model = '''
    Kind1 kind1 {
        a b c
    }
    Kind2 kind2 {
        a b c
    }

    3+6*(7+2*kind1.a); 4+5*(1+2*kind2.a); kind1.b;
    '''
    mm.model_from_str(model)
```

goto40:
I need some time to read #111 .
But I think it is mandatory to separate
 * global scoping (FQN, PlainName) ==> lookup based on natural model structure
 * local scoping ("RelativeName", "RelativeNameExtended") ==> lookup custom and relative, controlled by domain logic, not natural model structure (compare: the new Testcase model example)

Conceptually I do not yet see how to blend them...

docs/scoping.md
goto40: I add the renaming of "RelativeName" to #111

goto40:
I did not adapt scoping.md... Maybe we leave it that way (it uses the deleted examples)? Moved to #111

goto40:
I propose to postpone this pull request until #111 is discussed.

igordejanovic:
> I propose to postpone this pull request until #111 is discussed.

I agree. This requires some more discussion and design considerations.

goto40:
For me, this PR could be dropped (it will still be available as "closed" but not "merged"?).
We could maybe put a comment in issue 107 or so...

igordejanovic:
I agree. This issue should hopefully be fixed properly by #111 
Here is some valuable discussion so I'm mentioning #107 to have a link there and also closing.
Thanks for your work. :)

PR #112, opened 2018-11-17
Feature (request): uniqueness checker

dkrikun:
I've noticed that I am writing very similar processors to check for multiple definitions of objects (entities, properties etc.) in a DSL similar to entity example.
As there is a mechanism to help with name resolution, I think it could be useful to have one to check for multiple definitions.
I have implemented a simplistic checker as an object processor. It is limited for checking non-unique children by name. Imports and scoping are not taken into account.
Please advise how this should be improved.


dkrikun:
Oops this contains previous commits by @goto40

igordejanovic:
@dkrikun You could use `git rebase` to rebase your commits on the current master branch.

Issue #111, opened 2018-11-17
TEP-001: Reference resolving support in the textX language


igordejanovic:
This is an initial proposal for making an extensions to textX language to support resolving/scoping definition declaratively in the grammar.

textX support automatic reference resolving by [referencing grammar rule inside `[]`](http://www.igordejanovic.net/textX/stable/grammar/#references). Up until version ... the reference mechanism was very simplistic. The object was searched in the global model scope by the matched name and the referenced type. 

Starting from version 1.7.0, a new elaborate and flexible way to specify scoping rules was introduced thanks to the work done by @goto40. I think it is a very important part of language specification so I propose a change to textX grammar language to enable specification of scoping/ref. resolving rules in the grammar. The in-code mechanism should still remain to support some more advanced specification that can't be covered by grammar based approach. The idea is to make common things simple and specific/advanced things possible.

I propose to extend the `[]` ref style with optional additional expression that represent the search path.

For example:
```
myref = [OtherRule|FQN|a.b.c]
```
Where `a.b.c` would be a path to search. In this case it would be treated as absolute path starting from model root. Relative paths could be supported by a dot prefix `.a.b.c` and the semantic would be to start from the object where the name has been matched and follow the path. A special keyword `parent` could be used to navigate to the parent object or (as already introduced in scoping), `parent(TYPE)` would search up the parent chain until a particular type is found.

Resolving semantics could go like this:

Take this example from #107 as an example:
```
Model:
    kinds += GroupKind
    values += [LiteralKind|FQN|.kinds.vars]
;

LiteralKind : name=ID;
GroupKind:
    kindName=ID name=ID "{"
        vars *= LiteralKind
    "}"
;

FQN : ID'.'ID;
```
and the model:
```
Kind1 kind1 {
    a b c
}
Kind2 kind2 {
    a b c
}

kind1.a kind1.b kind2.a
```

To resolve `kind1.a` which is matched by FQN we take resolving specification `.kinds.vars` and start searching relatively from the `Model` instance (in this case absolute search expression `kinds.vars` would be equivalent). We split expression end evaluate `kinds` in the scope of the `Model` instance. We see it is an attribute of type `list` and then we use the first part of the matched FQN name to search by name in the list. This will result in `GroupKind` object `kind1`. We then evaluate the rest of the search expression in the scope of `kind1`. That would be a `vars` which is again a list so we use the rest of the FQN to search in this list yielding the object `a` of type `LiteralKind` which is found in the `kind1` instance.

[...]

PR #113, opened 2018-11-17
Feature (request): uniqueness checker

dkrikun:
I've noticed that I am writing very similar processors to check for multiple definitions of objects (entities, properties etc.) in a DSL similar to entity example.
As there is a mechanism to help with name resolution, I think it could be useful to have one to check for multiple definitions.
I have implemented a simplistic checker as an object processor. It is limited for checking non-unique children by name. Imports and scoping are not taken into account.
Please advise how this should be improved.


goto40:
Thank you for your work. 

I like your idea to add a library of standard object processors. Your proposal could be a starting point. Good idea!  

Discussion:

  * how do we organize the code? A validation module?

  * do we need more expressiveness (e.g., for an IDE) to report not only errors but also warnings? This involves probably fundamental API changes, and is maybe too much for now... 

  * For the concrete validator proposed here: can it be employed in a more generic way (without explicitly selecting entities or properties - maybe I did not see that, yet). 

  * we need to sort out the impact on scoping... 

Wait for more feedback... Best wishes!

igordejanovic:
> how do we organize the code? A validation module?

Yes, I vote for `validation` module with all generic/reusable semantic checks.

> do we need more expressiveness (e.g., for an IDE) to report not only errors but also warnings? This involves probably fundamental API changes, and is maybe too much for now...

We could think of mechanism to collect warnings on the model that could be utilized by IDEs. I think that even semantic errors could be collected as we could find all in one go (parsing phase is already over when we get to obj. processors). Yup, some of these are backward incompatible so we should make a branch for 2.0 development for all backward incompatible changes.

> For the concrete validator proposed here: can it be employed in a more generic way (without explicitly selecting entities or properties - maybe I did not see that, yet).

In my opinion if we have a collection of nameable things (and thus referencable) we should by default check for uniqueness. In this case `entities` and `properties` would be checked for uniqueness as they have name. I fail to see a use-case where this would not be right solution. Other way around is not that straightforward. If we have many collection of the same type, sometime they should be unique in between but more often not (due to scoping).

dkrikun:
> We could think of mechanism to collect warnings on the model that could be utilized by IDEs. I think that even semantic errors could be collected as we could find all in one go (parsing phase is already over when we get to obj. processors). Yup, some of these are backward incompatible so we should make a branch for 2.0 development for all backward incompatible changes.

I wanted to ask about it in another thread, but since it came up here: from what I understand all errors are non-recoverable because once an error is detected an exception is raised and thus the parser cannot continue parsing from where it stopped. Thus there are no "syntax warnings" and the parser cannot report more than one syntax error at a time, is that right?

> In my opinion if we have a collection of nameable things (and thus referencable) we should by default check for uniqueness. In this case `entities` and `properties` would be checked for uniqueness as they have name. I fail to see a use-case where this would not be right solution. Other way around is not that straightforward. If we have many collection of the same type, sometime they should be unique in between but more often not (due to scoping).

1. In what scope the objects should be checked for uniqueness? It makes sense to check in the parent object container, but for plain name import scope provider, objects at model top-level need be checked against objects at top-level in the "imported models", needn't they? Also it might be the case that one object is allowed to "shadow" or "rebind" a previously defined object with the same name. Another question, what about imported models, should their imporURI be checked for uniqueness (it makes some sense to)?

2. I am willing to work on this feature, could you outline some steps this should be broken into?

goto40:
@igordejanovic  @dkrikun : To be discussed.

Thank you @dkrikun for the detailed questions. I think your ideas lead us to a quite complete and flexible implementation of the checker. I tried to summarize your ideas...

### Uniqueness Checkers

I think you mentions the right aspects. We could add two checkers:
 * A **plain name** checker
 * A **FQN name** checker

Both of them should have several options (note type==rule):
 * a flag to perform **checks** on a **per-type** basis (same name, but different type allowed) or **globally** (no name is allowed twice).
 * an optional **white list** of types to check.
 * an optional **black list** of type not to check.
 * a flag to indicate if uniqueness is **per file** or **across files** (multi file models, visible models affected).

Note: a checker for, e.g., uniqueness of attributes within a class (you mentioned to check for uniqueness in the parent of a container) is - in my eyes - covered by an FQN uniqueness checker restricted to a white list indicating the attribute-type.

### Validator API

I suggest to stick to execptions to be thrown by the validators. We should **move the new warning/error API to a new issue and postpone it**. I could imagine that this design decision is quiet cheap to modifiy later.

dkrikun:
I may be messing the concepts, but what I meant is, for example, if the
grammar is:

```
Model:
    zoo1 *= Animal ';'
    zoo2 *= Animal
```
then the uniqueness should be checked separately in zoo1 and zoo2.

goto40:
Ah: I see.
 * So you suggest to allow a "Tiger" to be present twice (in zoo1 and zoo2)
 * In this case a FQN link will not be able to distibuish between them, since zoo1 and zoo2 have no names (even that the new checker would allow these double usage).
 
Do you think this is a use case? 

If you have
```
Model: zoos+=Zoo;
Zoo: name=ID animals+=Animals;
```
then a "normal" FQN checker would work...

dkrikun:
Yes, this the example, I am not 100% one needs to build it this way, and
adding name to zoo definitely works.

So the plain checker will only need to loop over each object's child
containers and check uniqueness per container.

I guess it won't be useful to check in non-container types, e.g.:

```
Model:
    alpha=Animal ';'
    zoo*=Animal
;
```
Then if the same tiger is both in the zoo and he's the alpha it is fine.

goto40:
I am not 100% convinced:
The plain name checker should check the uniquness of _all_ elements (maybe per type or independent of the type) - and not the uniqueness _per container_.

(if you wish the per-container, you use the FQN uniqueness, maybe restricted to animals if you wish - this will do the job).

The alpha-Tiger is not an issue to me:

```
Model:
  alpha = [Animal] ";"
  zoo += Animal;
```

The one-and-only tiger is the alpha! ;-)

goto40:
Note: FQN-uniqueness will distiniguish zoo_munich.tiger from zoo_paris.tiger
--> Thus, you will get uniqueness per-container (the container needs to have a name...)

If you need the uniqueness of elements in unnamed containers, we may need another case... (Maybe a UniqueNamesPerList("Rule.list_attribute").

dkrikun:
I see, so you might have tigers all over the place, in zoos, etc. -- as long as the same Tiger object name is used more than once in a model (or over multiple models, s.t. to options) it is considered non-unique.
On the other hand, FQN-uniqueness takes into account names of all ancestors up until the model root.

Regarding the unnamed containers I am now convinced that it is not so useful, because if the containers are unnamed, how one would address their elements in the first place? Somewhere in the FQN chain something must be different.

goto40:
Let's sleep one night. Maybe we could add some more positive /negative examples (like the ones in your first commits).

This may help illustrating all combinations we mentioned above... 

I am quite happy with our discussion, since we worked out some nice aspects... Good work! Thank you very much!

igordejanovic:
> I wanted to ask about it in another thread, but since it came up here: from what I understand all errors are non-recoverable because once an error is detected an exception is raised and thus the parser cannot continue parsing from where it stopped. Thus there are no "syntax warnings" and the parser cannot report more than one syntax error at a time, is that right?

You are right. There is no error recovery in the underlying Arpeggio parser so
the parser will raise an exception on syntax error. But, once the parsing is
done we can make all semantic checks in one go.

> In what scope the objects should be checked for uniqueness? It makes sense to check in the parent object container, but for plain name import scope provider, objects at model top-level need be checked against objects at top-level in the "imported models", needn't they? Also it might be the case that one object is allowed to "shadow" or "rebind" a previously defined object with the same name. Another question, what about imported models, should their imporURI be checked for uniqueness (it makes some sense to)?

This is a good question. Thanks. I guess that the scope we are looking for is
the one that is used for referencing. So we should analyze all references and
decide what is the proper scope for uniqueness of the referenced objects.
"shadowing" and "rebinding" are indeed interesting cases to discuss. I guess we
could automatically detect shadowing/rebinding but should have a way to specify
if it is allowed in the language or error/warning should be raised.

I guess we'll have a better picture once #111 is resolved so I suggest to first deal with that.

goto40:
I would leave this PR open, or reopen it for 2.0.0.

igordejanovic:
> I would leave this PR open, or reopen it for 2.0.0.

I agree to leave this open for now.

Issue #117, opened 2018-11-19
get_children_of_type: AttributeError: 'str' object has no attribute '__name__'


mrkafk:
Parsed string:

```
host example1 {
 hardware ethernet aa:bb:ff:20:fa:12;
 fixed-address 192.168.1.181;
 option host-name "example1";
 ddns-hostname "example1";
}

host example2 {
 hardware ethernet aa:bb:ff:20:fa:13;
 fixed-address 192.168.1.191;
 option host-name "example2";
 ddns-hostname "example2";
}
```

Code:
```
def get_hosts(s):
    grammar = """
    config: hosts*=host ;

    host: 'host' hostname=ID '{'
        'hardware ethernet' hardware_ethernet=/[0-9a-fA-F:]+/';'
        'fixed-address' fixed_address=/([0-9]{1,3}\.){3}[0-9]{1,3}/';'
        'option host-name' option_host_name=STRING';'
        'ddns-hostname' ddns_hostname=STRING';'
    '}'
    ;
    """
    mm = metamodel_from_str(grammar)
    model = mm.model_from_str(s)
    hosts = get_children_of_type('host', model)
```

Exception:
```
  File "./prog.py", line 58, in get_hosts
    hosts = get_children_of_type('host', model)
  File "/ve/local/lib/python2.7/site-packages/textx/model.py", line 125, in get_children_of_type
    typ = typ.__name__
AttributeError: 'str' object has no attribute '__name__'
```

Per `get_children_of_type` docstring, `typ` should be allowed to be a string?

goto40:
I checked your example with python 3.6 and the current textX version. I do not get the effect.

Which version of textX and python do you use?

```
from textx import metamodel_from_str, get_children_of_type

grammar = """
config: hosts*=host ;

host: 'host' hostname=ID '{'
    'hardware ethernet' hardware_ethernet=/[0-9a-fA-F:]+/';'
    'fixed-address' fixed_address=/([0-9]{1,3}\.){3}[0-9]{1,3}/';'
    'option host-name' option_host_name=STRING';'
    'ddns-hostname' ddns_hostname=STRING';'
'}'
;
"""
mm = metamodel_from_str(grammar)
model = mm.model_from_str('''
host example1 {
 hardware ethernet aa:bb:ff:20:fa:12;
 fixed-address 192.168.1.181;
 option host-name "example1";
 ddns-hostname "example1";
}

host example2 {
 hardware ethernet aa:bb:ff:20:fa:13;
[...]

Issue #105, opened 2018-11-14
[Python2] Trying to export a non-unicode string causes an error


imaduro:
Exporting a parsed model that contains non-unicode strings causes the following error on Python2

```
>>> my_string = "value + value2"
>>> parsed = example_meta.model_from_str(my_string)
>>> from textx.export import model_export
>>> model_export(parsed, "my_file.dot")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 163, in model_export
    model_export_to_file(f, model, repo)
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 267, in model_export_to_file
    _export(model)
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 222, in _export
    _export(list_obj)
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 195, in _export
    for attr_name, attr in obj_cls._tx_attrs.items():
AttributeError: type object 'str' has no attribute '_tx_attrs'
```

igordejanovic:
Related to #99 
We should have a better error report if a user pass a non-unicode object to `*_from_str`.

igordejanovic:
As of #120 textX strictly requires unicode.

PR #99, opened 2018-11-12
Fixes export failures on nested parses

imaduro:
While using Python2, if a grammar can have multiple nested expressions, the exporter will fail to export these if Strings are present within them. 

Example.tx
```
Expression:
    ArithmeticExpression | SubExpression
;

ArithmeticExpression:
    operand=Operand (operator=Operator operand=Operand)+
;

SubExpression:
    leftParen="(" sub=Expression rightParen=")"
;

Operand:
    ID | INT | SubExpression
;

Operator:
    "+" | "-"
;
```

then using the string `(value + (exp + bla)) - blabla`

```
Python 2.7.15 (default, May  1 2018, 16:44:08)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from textx import metamodel_from_file
>>> example_meta = metamodel_from_file('example.tx')
>>> my_string = "(value + (exp + bla)) - blabla"
>>> parsed = example_meta.model_from_str(my_string)
>>> from textx.export import model_export
>>> model_export("./", "my_file.dot")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 163, in model_export
    model_export_to_file(f, model, repo)
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 266, in model_export_to_file
    _export(model)
  File "/usr/local/lib/python2.7/site-packages/textx/export.py", line 194, in _export
    for attr_name, attr in obj_cls._tx_attrs.items():
AttributeError: type object 'str' has no attribute '_tx_attrs'
```

This bug only affects Python 2, Python 3 does not run into this error.

PR #120, opened 2018-11-21
Strict unicode requirement for (meta)model_from_str

igordejanovic:
Unicode requirement for (meta)-model strings API parameters made strict.
This should prevent common errors with Python 2.x

Related to #105 #99 #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@
    - https://github.com/igordejanovic/textX/pull/93
      - Changed attribute name for the metamodel object (from 
        "metamodel._parser" to "metamodel._parser_blueprint").
+   - https://github.com/igordejanovic/textX/pull/120
+     Unicode requirement for (meta)-model strings API parameters made strict.
+     This should prevent common errors with Python 2.x
+     See:
+     https://github.com/igordejanovic/textX/issues/105
+     https://github.com/igordejanovic/textX/pull/99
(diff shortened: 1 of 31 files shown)


goto40:
Very good! Maybe have a look at the inline comment. I possibly have overseen a point, but I think with minimal effort we could enhance the test. Else: perfect.

igordejanovic:
Thanks! What inline comment your are referring to?

goto40:
I was referring to https://github.com/igordejanovic/textX/pull/120 ("Should
we also have a test showing that an exception is raised, when 'str')...

Maybe I did not use the github ui correctly...


Am Do., 22. Nov. 2018, 10:49 hat Igor Dejanović <notifications@github.com>
geschrieben:

igordejanovic:
> I was referring to #120 ("Should we also have a test showing that an exception is raised, when 'str')... 

Er, sorry I still don't get it. Could you provide a link?

goto40:
Sorry. I just repeat the comment for file  tests/functional/test_strict_unicode.py:

Should we also have a test showing that an exception is raised, when 'str' in python 2 is used (without from future import...)... Maybe 'if python 2 then... '?

igordejanovic:
Ah, OK, no I don't think it would be necessary. The provided test covers passing in anything that is not a `unicode` on py2 or `str` on py3 I guess as we are using `text` "type" which is an alias for `unicode`/`str`.

Issue #123, opened 2018-11-22
Inspecting object types


dkrikun:
In robot example, `__class__.__name__` is used to discriminate between `InitialCommand` and `MoveCommand`. However, using this technique doesn't work when the objects being inspected are of abstract type, because `__class__.__name__` returns only the most derived type.

Is there a better way to check types of objects in the model?
On the related note, I'am considering to monkey-patch the objects in the model, specifically to add a method which will polymorphically do something different depending on the object type.

Is it a good idea?

igordejanovic:
Hm, there is `textx_isinstance` in the [scoping.tools](https://github.com/igordejanovic/textX/blob/master/textx/scoping/tools.py#L35) done by @goto40. Haven't tested is it working in this case. It looks like it should.

> On the related note, I'am considering to monkey-patch the objects in the model, specifically to add a method which will polymorphically do something different depending on the object type.
Is it a good idea?

Well, monkey-patching is almost never a good idea but sometimes it is the most pragmatic solution. In this case [custom classes](http://www.igordejanovic.net/textX/latest/metamodel/#custom-classes) might solve your problem.

dkrikun:
The problem with `scoping.tools` is how does one gets a grip on a textx
class in the first place?

igordejanovic:
You can get a class from the meta-model. That part is not covered by the docs but meta-model object acts as a dict which you can query by class/rule name and get that object.

igordejanovic:
@goto40 BTW, `textx_isinstance` and some others generally usable function should probably be moved out of scoping module and exported/documented as a public API.

goto40:
There is a (very simple) unittest. The first test in https://github.com/igordejanovic/textX/blob/master/tests/functional/test_scoping/test_scoping_tools.py

If you have problems I can enhance the test...

dkrikun:
> > On the related note, I'am considering to monkey-patch the objects in the model, specifically to add a method which will polymorphically do something different depending on the object type.
> > Is it a good idea?
> 
> Well, monkey-patching is almost never a good idea but sometimes it is the most pragmatic solution. In this case [custom classes](http://www.igordejanovic.net/textX/latest/metamodel/#custom-classes) might solve your problem.

I have looked at the expression examples in the repository. There, you have defined custom classes for all (or almost all) elements of the grammar so that you can have `value` or `op` properties defined for various elements.
The problem with this approach is that it negates the convenience for having textx to define model classes for you.
[...]

PR #126, opened 2018-11-23
Dynamically add properties to object classes

dkrikun:
Add a version of calc.py in expression example that exercises
dynamically adding properties to object classes.

See #123


igordejanovic:
Looks good. Thanks.

PR #127, opened 2018-11-23
Scoping typos docs

igordejanovic:
These are some fixes and additions I've done during reading of scoping docs and implementation.

### DIFF
--- a/textx/scoping/__init__.py
+++ b/textx/scoping/__init__.py
@@ -93,7 +93,7 @@ class ModelRepository(object):
         self.filename_to_model = {}
 
     def has_model(self, filename):
-        return filename in self.filename_to_model.keys()
+        return filename in self.filename_to_model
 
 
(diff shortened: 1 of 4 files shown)


textx/scoping/__init__.py
```diff
@@ -93,7 +93,7 @@ def __init__(self):
         self.filename_to_model = {}
 
     def has_model(self, filename):
-        return filename in self.filename_to_model.keys()
+        return filename in self.filename_to_model
```
goto40, line 96: Is this the same or not (old code/new code)?

goto40 (approved):
Very good! Thank you.

Ready to merge...

textx/scoping/__init__.py
igordejanovic: don't understand the question. dict uses keys when iterating over it or with `in` operator so no need to explicitly call `keys()`.

PR #128, opened 2018-11-23
Add calc_isinstance.py example

dkrikun:
The example exercises usage of `text.scoping.tools.textx_isinstance()` to inspect model objects types during traversal.

See #123


dkrikun:
Question: while experimenting with this, I've tried to rewrite
```Operand: PrimitiveOperand | CompoundOperand;```
to be directly:
```Operand: PrimitiveOperand | ('(' Expression ')');```

But the somehow '(' is attempted to parse as 'Factor' op attribute.
Why is that? Why isn't it consumed by '(' around Expression in the first place?

**Comment by igordejanovic** [comment]
Looks good. Could you please update README for the expression examples for this and monkey patch example.

> `Operand: PrimitiveOperand | ('(' Expression ')');`
> But the somehow '(' is attempted to parse as 'Factor' op attribute.
Why is that? Why isn't it consumed by '(' around Expression in the first place?

Not sure why is that. Could you make a PR with that version to investigate?

**Comment by dkrikun** [comment]
Can I make the change on this PR?

**Comment by igordejanovic** [comment]
Sure.

**Comment by dkrikun** [comment]
Any change you had a look on this problem? Thanks

**Comment by igordejanovic** [comment]
Sorry, haven't had a chance. Will take a look during the weekend.

**Comment by goto40** [comment]
I had a quick look:
 * you can debug the code, but it is painful to click through the deeply nested model. I used the following snipet to print the model.
```
def pprint(m,i=0,aname=""):
    if aname.startswith("_tx"):
        return
    print(" "*i+"->"+aname+":"+str(m))
    if isinstance(m, list):
        for c in m:
            pprint(c, i + 2, "("+aname+")")
    elif hasattr(m,"_tx_metamodel"):
        for k,c in m.__dict__.items():
            if k!="parent":
                pprint(c,i+2,k)
```
 * I checked that the Operand is a base of Expression:
```
    o=calc_mm['Expression']()
    assert textx_isinstance(o, calc_mm['Operand'])
```
 * *However*, It seems that op="(" when op should be an Expression... Somehow <"(" Expression ")"> is transformed into a string "("...

According to http://www.igordejanovic.net/textX/stable/grammar/#rule-types (as far as I understand...), Operand should be an abstract rule... Maybe @igordejanovic has the final hint next weekend.

igordejanovic:
Thanks @goto40 for the analysis, it was very helpful. 

The problem was not handling properly an alternative which contains str/regex matches in abstract rules. We haven't had that situation before but it's generally usable.

The fix is in PR #134

igordejanovic:
This should work with the current master now and is almost ready to be merged.

@dkrikun Please clean up the branch (I think this could all be just a single commit, or if you have better idea do as you please), add `from __future__ import unicode_literals` to the `calc_isinstance.py` and rebase on the current master. Thanks.

PR #130, opened 2018-11-24
Simplify usage of flake8. Configure using `setup.cfg`.

igordejanovic:
This change simplifies flake8 running from command line.


goto40:
Great!
Much better than before.

goto40 (approved):

PR #132, opened 2018-12-03
added missing package

goto40:
With my new linux installation I reconfigured by venv... Here I had to add one package.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -4,3 +4,4 @@ coverage
 jinja2
 mkdocs
 mike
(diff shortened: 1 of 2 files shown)


igordejanovic:
I think it is only needed for performance tests. Those requirements are [here](https://github.com/igordejanovic/textX/blob/58d4c47d821f13abde9311dc1a28969b779f74c7/tests/perf/requirements.txt)
Am I wrong?

goto40:
Sorry. My bad.

Issue #103, opened 2018-11-14
Import like in python?


dkrikun:
Hello,

I would like to have an import mechanics similar to python. In particular, I would like to have the imported module name act as a namespace. For example:

```
# imported module 'base.mod'
class a {};
```

```
# main module
import 'base.mod'
var x = new base.a()
```
So I've put the 'ImportFQN' provider for '*.*', but the name is still resolved as 'a' and not 'base.a'.
I do not understand what the FQN providers consider for valid prefixes when resolving a name.

Related to this, could be nice to be able to do something like `import x.y.z as w`, is it possible?

Thanks :smile:

goto40:
At the moment all includes are flat. The current importuri scope providers
are not capable of fulfilling your needs.

I see different possibilities of introducing your requirements into the
scope provider design. I would like to take my time to write them down and
to discuss them.

Thank you for the idea.

Note: the rule containing the importuri field contains also the imported
model. With a custom scope provider, this should allow to implement an
appropriate scope provider. See:
http://www.igordejanovic.net/textX/stable/scoping/ ("Model objects formed
by the rules with an importURI attribute get an additional attribute
_tx_loaded_models which is a list of the loaded models by this rule
instance.").

(to be continued)

Am Mi., 14. Nov. 2018, 09:48 hat Daniel Krikun <notifications@github.com>
geschrieben:

goto40:
I think it is not so difficult. As mentioned in the note above, nearly all information is available. I will setup a unittest to clarify the needs and make a small demo. We can discuss it then...

In particular:
  * Import "folder_a/folder_b/file.ext" --> import folder_a.folder_b.file (hook to transform an importUri string to a filename)
  * import "file" as name (references relative to the import object)

goto40:
Check #106
To be discussed... But I think your wishes were quite clear and the problem to be solved is unambiguous.

goto40:
(sorry - reopened, closed issue by accident)

goto40:
(not fully finished - I try your example)

goto40:
I integrated your example above (without the "import as" like in "import xy as ab"). 

"Import as" is covered in the unittests.

dkrikun:
awesome :blush:

PR #133, opened 2018-12-04
Pull request #114 addition

goto40:
Playing around for TEP-001 I refactored the solution for #103 to be more generic.

**Changes**
Instead of hard code the importURI/_tx_loaded_models attriute when resolving a FQN, I 
allowed to provide a scope_redirection_logic to replace a current object in an FQN chain.

**Effect**
With this, #103 is implemented. However it also allows to define package-aliases (see scoping.md and the regression test for #103).

**Docu**
This work also helped to **document, why the Postponed mechanism is not part of the FQN** (docu  adapted). 

Explanation text (no open point):
"It remains unclear if a processed unresolved element (Postponed) will be relevant for the current name. This **would** easily create circular dependencies for the current FQN implementation. Therefore, Postponed is excluded (rational included in scoping.md)."


tests/functional/regressions/test_issue103_python_like_import.py
```diff
@@ -50,3 +52,105 @@ def importURI_to_scope_name(import_obj):
...
+            package p2 {
+                class a {};
+            }
+        }
+        using p1.p2 as main
+        var x = new p1.p2.a()
+        var y = new main.a()
+    ''')
+
+    # first test (negativ example)
```
igordejanovic: typo: `negativ` -> `negative`

tests/functional/regressions/test_issue103_python_like_import.py
```diff
@@ -50,3 +52,105 @@ def importURI_to_scope_name(import_obj):
...
+                package p2 {
+                    class a {};
+                }
+            }
+            using p1.p2 as main
+            var x = new p1.p2.a()
+            var y = new error.a()
+        ''')
+
+    # first test (negativ example)
```
igordejanovic: typo: `negativ` -> `negative`

textx/scoping/providers.py
```diff
@@ -153,6 +162,18 @@ def _find_obj_fqn(p, fqn_name, cls):
             """
 
             def find_obj(parent, name):
+                if parent is not current_obj and \
```
igordejanovic, line 166: Do we need `current_obj` parameter in all these function. As I see it carries the value of top level `obj` and is never changed. Could we just say here `if parent is not obj and ...`  and remove all those `current_obj` params?

textx/scoping/providers.py
```diff
@@ -201,20 +218,20 @@ def _find_referenced_obj(p, name, cls):
...
+            ret = _find_obj_fqn(p, name, cls, current_obj)
             if ret:
                 return ret
             while hasattr(p, "parent"):
                 p = p.parent
-                ret = _find_obj_fqn(p, name, cls)
+                ret = _find_obj_fqn(p, name, cls, current_obj)
                 if ret:
                     return ret
             return None
```
igordejanovic: I've just noticed this. In Python if function don't return anything it is implied that it returns `None`.

textx/scoping/providers.py
```diff
@@ -201,20 +218,20 @@ def _find_referenced_obj(p, name, cls):
...
 
         from textx.model import ObjCrossRef
         assert type(obj_ref) is ObjCrossRef, type(obj_ref)
         obj_cls, obj_name = obj_ref.cls, obj_ref.obj_name
-        ret = _find_referenced_obj(obj, obj_name, obj_cls)
+        ret = _find_referenced_obj(obj, obj_name, obj_cls, obj)
         if ret:
             return ret
         else:
```
igordejanovic: Can we just say here `return _find_referenced_obj(obj, obj_name, obj_cls)` ? 
No need for `if` afterwards I would say.

igordejanovic (changes requested):
Looks good. Sorry for taking me so long. 
I have some minor changes to discuss. Please see inline comments.

goto40:
Thank you for the feedback. It is ok for to leave some pull requests aside until we have time to discuss (this is no central feature)... I will work on the comments now.

tests/functional/regressions/test_issue103_python_like_import.py
goto40: fixed

tests/functional/regressions/test_issue103_python_like_import.py
goto40: fixed

textx/scoping/providers.py
goto40: Absolutely! I removed the superfluous parameters. However I changed the top-level parameter name to "current_obj" to make the code more readable.

textx/scoping/providers.py
goto40: OK. Nice to know. I removed two "return None" but also added one comment at the end of the while loop indicating that an implicit "return None" is expected in some cases.

textx/scoping/providers.py
goto40: Ok. See last comment (this is the second return None which I removed).

goto40:
Changed code according to the comments.

goto40:
Note: I noticed that we can "Squash and merge" instead of "Merge pull request" in the github interface (you need to drop down the green button). This might make it easier for other commiters to just say "merge and squash" my stuff into  the master...

igordejanovic:
Thanks. Looks good. Merging.

> Note: I noticed that we can "Squash and merge" instead of "Merge pull request" in the github interface (you need to drop down the green button). This might make it easier for other commiters to just say "merge and squash" my stuff into the master...

Yes, that can be used if all commits should be squashed but if finer control is needed then git interactive rebase is the way to go.

PR #136, opened 2018-12-14
Add calc_isinstance.py example

dkrikun:
The example exercises usage of text.scoping.tools.textx_isinstance() to inspect model objects types during traversal.

See #123


igordejanovic:
Thanks.

Issue #140, opened 2018-12-19
Assertion error while calling object processors with multi meta models (extended grammars) and custom types


goto40:
I encountered a problem with
 * two grammars, one extending the other (the second grammar allows to instantiate elements 'inherited' from the other). In contrast to the "referenced_metamodels" feature, we get the meta classes created for both meta models separately, not reusing the already created meta classes (see https://github.com/igordejanovic/textX/blob/master/docs/multimetamodel.md). 
 * multi meta models (multiple model files including each other and referencing each other)
 * custom classes (passed to both meta models)
 * and object processors.

Grammar A.tx
```
Model: es+=E;
E: 'E' name=ID '{' c+=C '}';
C: C1;
C1: 'C1' name=ID t=TXT;

TXT: txt=STRING;
``` 

Grammar B.tx
```
import A
Model: x+=X;
X: 'X' name=ID '{' r+=R '}';
R: R1|C1;
R1: 'R1' ref=[E]; 
```

a custom object is created for C1:
```
class C1:
    def __init__(self, **kwargs):
        for k in kwargs.keys():
            setattr(self, k, kwargs[k])

# ...

    global_repo = scoping.GlobalModelRepository()
    repo = scoping_providers.PlainNameGlobalRepo()
    repo.register_models(os.path.dirname(__file__)+"/multi_obj_proc/*.a")

    mm_A = metamodel_from_file(os.path.join(
        os.path.dirname(__file__),
        "multi_obj_proc",
        "A.tx"
    ),global_repository=global_repo, classes=[C1])
    mm_B = metamodel_from_file(os.path.join(
        os.path.dirname(__file__),
        "multi_obj_proc",
        "B.tx"
    ), global_repository=global_repo, classes=[C1])

    mm_B.register_scope_providers({"*.*": repo})

    def proc(a):
        print(a)

    mm_A.register_obj_processors({"C1": proc})
    mm_B.register_obj_processors({"C1": proc})

    scoping.MetaModelProvider.clear()
    scoping.MetaModelProvider.add_metamodel("*.a",mm_A)
    scoping.MetaModelProvider.add_metamodel("*.b",mm_B)

    _ = mm_B.model_from_file(os.path.join(
        os.path.dirname(__file__),
        "multi_obj_proc",
        "b.b"
    ))
``` 

Model a.a
``` 
E name1 {
    C1 c1 'hello1'
}
E name2 {
    C1 c1 'hello2'
}
``` 
Model b.b
```
X n1 {
[...]

PR #141, opened 2018-12-19
Bugfix multimeta model

goto40:
I encountered a problem (#140).
 * It has to do with the fact that we have different instances of meta classes in two meta models based on a common grammar ("import grammar"-feature).
 * In the object processor handling function, we compare two base classes for equality (if "a is b"). This is changed in this proposal to a comparison of the fully qualified names of the meta classes.

Summary:
 * The fix is understood. 
 * The effect is not understood 100% (not clear under which circumstances exactly this happens).


igordejanovic:
Looks good. I would suggest moving test and associated files under `tests/functional/regressions/140/` to reduce the clutter.

This solves the issue at the moment but the cause of the problem seems to be that we have `TXT` created in both A and B meta-models. I'm not sure that we need both, probably should be created just in the source meta-model and used where necessary.

Those two seems to get confused while passing `metaattr.cls` on line 566.

I'm for merging this fix, once the tests are moved to a separate folder, and maybe leaving issue open until we fully investigate and understand the source of the problem.

goto40:
Thank you.
I changed the folder to "issue140", in line with the other folders in this directory.

igordejanovic:
Thanks. Looks good. Merging.

PR #151, opened 2018-12-24
Use Source Code Pro as font for <code> in docs

### LINKED ISSUE #143: Suggestion to change the current choice (`readthedocs`) for mkdocs theme for better readability
I am going through the tutorial now.

In the HelloWorld I mistake the comma for a period.

I like to recommend switching the theme to something that makes the code syntax highlighting more readable. 

Not sure if it's possible, but I saw a free theme for Guzzle http://www.writethedocs.org/guide/tools/sphinx-themes/#guzzle-theme

and I checked out their website, and at least the comma is quite obvious. 

http://docs.guzzlephp.org/en/stable/

What does the community think?

## Update

I think guzzle is a purely sphinx theme and does not have a mkdocs theme equivalent. So ignore my recommendations

simkimsia:
closed #143

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -1,2 +1,3 @@
 div.col-md-9 img[alt="textX logo"] { background-color: transparent; border: none; }
 .rst-content .section ul li p:last-child { margin-bottom: 12px; }
+code { font-family: 'Source Code Pro', monospace; }


igordejanovic:
Looks better. Thanks.

PR #153, opened 2018-12-24
Add instructions on running mkdocs locally

simkimsia:
for the ignorant. Ref https://github.com/igordejanovic/textX/issues/143#issuecomment-449556703

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,25 @@ textX could always use more documentation, whether as part of the
 official textX docs, in docstrings, or even on the web in blog posts,
 articles, and such.
 
+#### How to Test the Documentation Locally
+
+textX is currently using `mkdocs`, a documentation generator, to generate the docs into html files. 
+
+To test the docs locally, you need to follow the first 3 instructions at the [Get Started!](#get-started) section
+
+1. Fork the repo (one-time effort)
+2. Clone your fork locally (one-time effort)
+3. Create a virtualenv for the fork and install the relevant libraries (one-time effort)
+
+Once you complete the above 3 instructions, you can now:
+
+4. Activate the virtualenv
+5. Run `mkdocs serve` at the root folder
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -41,6 +41,25 @@ textX could always use more documentation, whether as part of the
...
+3. Create a virtualenv for the fork and install the relevant libraries (one-time effort)
+
+Once you complete the above 3 instructions, you can now:
+
+4. Activate the virtualenv
+5. Run `mkdocs serve` at the root folder
+
+`mkdocs` will run a webserver that serves the documentation at 127.0.0.1:8000
+
+To make changes to the configurations, you can look at `mkdocs.yml`. For more information on how to use mkdocs, visit this [site](https://www.mkdocs.org). Essentially, you do not need to install mkdocs as a software locally to test the documentations.
```
igordejanovic: Not sure what do you mean by `you do not need to install mkdocs as a software locally...`. You actually do need mkdocs. It is installed as a part of `pip install -r requirements_dev.txt`. See the file, there is `mkdocs` in it. I suggest to remove this sentence entirely as it is confusing.

igordejanovic (changes requested):
Looks good. Just one change is required at the end. Please, see below.

CONTRIBUTING.md
simkimsia: Understood. Will make another change and then squash the commits. I have to go get dinner. Then I will come back and fix this asap.

goto40 (commented):
Note : when serving locally some files get generated. Can you check .gitignore if the local docu is correctly ignored.

OK. Just checked. Seems to be OK.

CONTRIBUTING.md
igordejanovic: Sure. No rush :)

simkimsia: @igordejanovic done. I force pushed the squash. It should be updated soon.

igordejanovic:
Thanks. Merging.

Issue #155, opened 2019-01-13
Fool-safe handling of non-abstract rules without assignation


Aluriak:
Hi !

I (the fool in this situation) lost a lot of time figuring why my grammar was failing hard:

    import textx

    METAMODEL = textx.metamodel_from_str("""
    Model: expression=Expr ;

    ParenExpr: '(' Expr ')';
    Expr: ParenExpr | BaseExpr;
    BaseExpr: left=Condition operations*=Operation;
    Operation:  op=BoolOperator remaining=Condition;

    Condition: ParenExpr | RawCondition;
    RawCondition: id=Identifier op=MathOperator val=INT;

    Identifier: id=/[a-zA-Z0-9_-]+/;
    MathOperator:  op=/=|>|</;
    BoolOperator:  op=/AND|OR/;
    """)

    raw_model = METAMODEL.model_from_str('(b=3 OR c=4)')
    print(raw_model.expression)

    raw_model = METAMODEL.model_from_str('a=2 AND (b=3 OR c=4)')
    print(raw_model.expression)

On the input `a=2 AND (b=3 OR c=4)`, it would provides me correctly the python objects for `a=2`, but the string `(b=3ORc=4)` instead of a hierarchy of Expr objects for the parenthesed part.
The answer was in the `ParenExpr` rule that should write:

    ParenExpr: '(' expression=Expr ')';

I was fooled by the following writing:

    AbstractRule: ARule | AnotherRule;

Which do not require any assignation.

My take on this is that this behavior is totally unexpected. Is there any rationale on this ?
Else, i have an opinion to share:

I shouldn't be able to write `ParenExpr: '(' Expr ')';` if it yields such type-weird results, unless i explicitly ask for it with a `cast_to_str_when_no_parsing_target` argument.

A probably better fix would be to make such rule an abstract rule. The general definition of abstract rule would then be: *a rule having only one possible target* (because there is only fixed strings elsewhere, or because you have a choice).

In a more general way, i do think that making a grammar just to raise an error if it does not pass is a *special* case. The regular one is to use all parsed data.
As a consequence, i would prefer textx to (by default) throw an error to me when i'm writing something like `Rule: RuleA RuleB`, i.e. a rule that parse while forgetting the parsed data.

Obviously, i am far from expert in grammar related things, and i just happen to run into that problem in my particular case: maybe this is a well-though feature, not a design issue. Would love to hear about the rationale.

goto40:
Hi. Thank you for the example.

[...]

PR #159, opened 2019-01-17
Issue155 test code

goto40:
The example code of #155. 
I think it is a very good example (Thank you @Aluriak).


goto40:
Hmm... I think I max have reproduced the effect observed by @Aluriak for python 3.5. See Travis!

goto40:
It is unclear to me why it is not failing on my fork? (see next comment)

goto40:
After restarting the build for this pr on Travis manually, the test passed... Then again failed. Needs to be analyzed!

igordejanovic:
Hm, definitely, that non-determinism is disturbing. Might be an issue of using dicts and relying on items order?

Aluriak:
> Hm, definitely, that non-determinism is disturbing. Might be an issue of using dicts and relying on items order?

Could be, although dict order was introduced in 3.6 IIRC, and all samples were tested with 3.6+ on my side.

Also, if the theoretical framework should be determinist, then use of unordered dict shouldn't change anything. My point is, even if dict were unordered, you shouldn't have any non-deterministic behavior, unless you explicitely rely on dict order to encode some determinism-related behavior.

To test that, one funny way is to run all the tests, but with an added artificial mixing of dictionnaries (mix keys, rebuild a dict using that new key order). If a dict order is found important, then i see two solutions:

- indicate explicitely that the order in this dictionnary is necessary, and use instead `collections.OrderedDict` to ensure 3.5- compatibility
- store the order/determinism-related-behavior by other means

igordejanovic:
@Aluriak Sure. The underlying theoretical framework (PEG) is deterministic but implementation still doesn't have to be, which should be treated as a bug. Having test fails sometimes, and only on 3.5 is very curious. @goto40 definitely should be analyzed and the root case should be identified.

goto40:
Internal note: 
 * High level effect: The difference seems to happen with "process_node" in model.py (a lot of printf-debugging).
 * The difference emerges from ```METAMODEL["ParenExpr"]._tx_type``` being "abstract" one time and "match" another time.
 * \_\_iter\_\_ in metamodel.py depends on the dict order. Some experiments with sorted keys (locally in \_\_iter\_\_) did not yield the error, even when inverting the order (result not 100% sure, since only a few experiments were made)
 * in lang.py: "def _has_nonmatch_ref(rule):" can return None (actually happens; seems "normal"; should maybe be False in that case)
 * In "lang.py": "def _determine_rule_types(self, metamodel):" iterates over the metamodel entries. 

**Summary:**

This was an observation session (no commit/push). I just played with python 3.5 and a lot of printfs... I could reproduce the effect quite often (>1 effect per 10 runs).

The key problems seems to be ```METAMODEL["ParenExpr"]._tx_type``` being different in some cases, which  somehow causes the effect. Maybe ```_determine_rule_types(self, metamodel)``` in lang.py is a good starting point for a walkthrough.

igordejanovic:
@goto40 Thanks for the insights. Will give it a debugging try on 3.5 as soon as I find some time. It really bugs me why does it happen on this version only.

goto40:
As @Aluriak mentioned : they changed the dicts in 3.6. This is also what I
found on SO.

goto40:
A candidate for a quickfix could be, as noted above,  ```__iter__``` in metamodel.py
>>    * ```__iter__``` in metamodel.py depends on the dict order. **Some experiments with sorted keys (locally in __iter__) did not yield the error**, even when inverting the order (result not 100% sure, since only a few experiments were made)

 * But I prefer to have the root cause completely understood (instead of "just sorting the one or the other dict").
 * However, a quick-fix could help us to jump to 2.x.x. We could still create a new ticket or leave the relevant tickets open.

Issue #148, opened 2018-12-21
Generate DSL documentation from metamodel (grammar)


alensuljkanovic:
Once DSL is created, documentation still needs to be written manually. 

If somehow we could document elements inside grammar, documentation will be part of metamodel (?). This would allow us to generate documentation in different formats (markdown, xml, html, ...). Plus, I believe that it would be easier to maintain the documentation, since it's in the same file as a language definition.

An interesting paper on this topic can be found [here](https://www.sciencedirect.com/science/article/pii/S1477842417301811).

goto40:
Do you mean something like Javadoc for grammars...?  The referenced paper seems to describe another approach.

igordejanovic:
Sounds interesting. I think it is worth researching the possibilities.

I guess something like javadoc approach would be relatively easy to implement. We already have [textX grammar defined in textX](https://github.com/textX-tools/textX-languageserver/blob/master/src/metamodel/textx.tx). :)

alensuljkanovic:
@goto40 actually, yes. Support for comment blocks is already there, so they could be used as docstrings.

Aluriak:
Hello !

I want to tackle this problem, since i needed it few days ago, and will probably need it in the future.
I made a first proof of concept [here](https://github.com/aluriak/textx-dsldoc), showing that something is definitively possible (although it seems that most code is a hand-crafted way to explore the structures… is there any standard visitor implementation ?).

I would like to open the discussion about (1) integration into textx and (2) methods to explore the grammar/parser i do not know about.

End goal is obviously to get a good doc generator.

goto40:
Great!

I need some time to look at it. 

Quick idea: Showing also the type of the rule (e.g. abstract) would greatly improve the overall compréhension of a grammar. You probably already integrated this idea.

Aluriak:
Take your time ; maybe just take a quick look to the output html, see what information is missing, that would make for a first feedback.

How can i detect the abstract type ?

igordejanovic:
@Aluriak Looks really good so far! It would be really nice to integrate it as a part of `textx` command. I guess the best way for this and all further sub-commands would be to have an extensibility mechanism in place (something very similar to what is implemented in [textx-tools](https://github.com/textX/textx-tools/blob/master/setup.py#L59)). In that way all additional sub-command could be an optional dependencies of textX and if installed via pip would be visible and accessible.

@goto40 
[...]

PR #162, opened 2019-01-25
CLI extensibility

### LINKED ISSUE #161: A way to integrate new textX CLI subcommands
Related to #148 #42 

See the initial discussion [here](https://github.com/textX/textX/issues/148#issuecomment-456643632).

The idea would be to provide an extension point for new sub-commands. In that way `textx` cli could be extended by new sub-commands by both official and un-official projects installable using `pip`.

See how it is done in the [textx-tools](https://github.com/textX/textx-tools/blob/master/setup.py#L59) project with `list-gens` and `list-langs` commands. For example [this](https://github.com/textX/textx-tools/blob/master/txtools/gen.py#L12).

igordejanovic:
closes #161 

This PR:
- migrates CLI to the [click](https://github.com/pallets/click) library,
- implements CLI extensibility. Other Python project may now add `textx` sub-commands and sub-command groups. Those are dynamically discovered. It is enough for the project to register extension point and be installed via `pip`. ~~This PR will be open until the docs are written~~. For now please see the test.

@goto40 @danixeee Please test/review.


igordejanovic:
Related to #148 
@Aluriak Please test if this feature will be what is needed to integrate doc generator with the textX cli.

goto40:
Do I understand correctly, that the purpose is to allow meta modeling tools to register? Like a generic model pretty printer, a meta model docu generator, etc.?

igordejanovic:
Yes, instead of each tool having its own CLI command and forcing user to remember a bunch of CLI commands this enable tools to register their CLI commands under `textx` command so the user just use `textx` to query and invoke commands.

You can test this by installing textX from this branch source with `pip install -e`, than install the example project from the test `subcommands/example_project`. When you call `textx` you will see now that there is one new subcommand and one subcommand group registered which you can invoke just like you would invoke built-in commands (`check` and `visualize`).

danixeee:
Looks good to me. I have successfully added `textxls` command group to the textX CLI and it is working fine. Here are `textx-ls-core` [entry points](https://github.com/textX/textX-LS/blob/danixeee/textxfile-lang-support/textX-LS/core/setup.py#L39).

Model validation for `Textxfile` with `textxls` CLI:
image

goto40:
I will try to add a Plantuml export, but I need some time (5-10 days). You can proceed and close if you need to - and we can reopen in cases of problems...

igordejanovic:
Great!

@danixeee I would suggest group name to be just `ls`. No need for a `textx` prefix.

@goto40 Sure, take your time. Is your idea to make separately installable tool to provide this option or to make it a part of the `textx` iteself? Maybe we could have it as an option on `visualize` command (e.g. `dot` being the default and `plantuml` being optional?
Ok, I'm merging this PR and if there is any issue we'll reopen.

igordejanovic:
I have merged this PR and closed #161 in order to have it more accessible for ongoing tools development. The docs will be tracked on #163.

igordejanovic:
Added the [docs on CLI extensibility](http://textx.github.io/textX/latest/textx_command/#extending-textx-command).

Aluriak:
Sorry for late reply, i had a lot on my plate lastly. I am currently working on this. Works well, but some issues are incoming.

PR #165, opened 2019-01-29
Plantuml

goto40:
* Started to integrate a **plantuml output** variant: at the moment for meta models only (not for models).
 * Integrated as **optional choice to the visualize command (click)**.
 * Limitations (already existing before this PR): the "referenced_metamodels" feature of metamodels is not compatible with the command line version of the export feature (because the referenced_metamodels-information is missing). 

Tests/Activities: I **refactored the existing dot export** and used a render_all_grammars script (see below) to compare (diff) the new and the old behavior. Only visual checks where possible, since ids are not stable.

An **example script is created to build all grammars** found as *.tx files in the textX project. The generated output is included here: https://github.com/goto40/textX/tree/plantuml/examples/render_all_grammars

Found **issues**:
 * DOT export effect: **number instead of a name** (also happens in master): https://github.com/goto40/textX/tree/plantuml/examples/render_all_grammars#first_diamondtx-1 
 * DOT/PLANTUML **transitive grammar imports not rendered?** (also happens in master): https://github.com/goto40/textX/tree/plantuml/examples/render_all_grammars#firsttx-2
   * Here, I miss "fourth.Second" visible in https://github.com/goto40/textX/tree/plantuml/examples/render_all_grammars#thirdtx-4 (not to be confused with second.Second).

**Question:**
 * Is this plantuml-export option something we want as a core-part of textx, or should we move (maybe the whole export stuff) to a separate project.
 * Is the simplified non-common-rule handling for plantuml (compared to dot) ok? We just output the type (as UML stereotype). I think I need to do something for this ("Value" in https://github.com/goto40/textX/tree/plantuml/examples/render_all_grammars#rhapsodytx) --> Maybe add a UML "inheritance" relation?


igordejanovic:
@goto40 Looks great! I'll review it this weekend.

Yeah, I was thinking also to remove all CLI tooling from the core library and provide it as a separately installable package, when I started textx-tools project, but have backed off from that idea as maybe too disruptive for existing users. Maybe we should reconsider it as it makes sense to have a core lib that is the only package needed in runtime/deployment and add-on package which provides CLI/dev stuff like  visualizers, checkers etc. needed only during development. That way we could keep run-time dependencies low while still being able to provide more features for the developers, and not being reluctant to introduce new dependencies if that would make dev. job easier.

For the non-common rules support, UML is more constrained than what can be displayed using dot, so I guess it is perfectly fine to go like this.

BTW, the script for generating outputs from all .tx files is usable for testing but we shouldn't check into repository all those generated files.

goto40:
I will definitively remove the output after review...

igordejanovic:
@goto40 Great work. I like the new approach using renderers.

One suggestion: Since PlantUML is a nice human-readable DSL I suggest for generated code to be nicely layout, i.e. the header should be aligned to first column, each class should have 2 empty lines before and all class attributes should be indented.

Please register the issues you have found as a separate GitHub issues so we can track and deal with those later.

`render_all_grammars` script looks very useful for a project with many textX grammars. You might consider making a new tool that extends textX command (with recently introduced extensibility) with the new command that implements this feature.

Please add/update the `visualization` page in the docs to make notion about new visualization format. Maybe we could put both dot and pu images in the [Meta-model visualization](http://textx.github.io/textX/stable/visualization/#meta-model-visualization) to show the difference in rendering. Furthermore, we should make a note that PU is not available for model rendering at the moment. Also, we should state at the beginning of the Visualization section what that PlantUML could be used and what is needed to be installed in that case.

goto40:
Thank you for the feedback. I will continue soon...

goto40:
(Work in progress)

- [X]  nicely layout, i.e. the header should be aligned to first column
- [x] register the issues you have found as a separate GitHub issues so we can track and deal with those later.
- [x] `render_all_grammars` -> new tool that extends textX command: N/A (later).
- [x] add/update the `visualization` page in the docs... 
  - [x] Furthermore, we should make a note that PU is not available for model rendering at the moment. ... 
  - [x] Also, we should state at the beginning of the Visualization section what that PlantUML could be used and what is needed to be installed in that case.

goto40:
> `render_all_grammars` script looks very useful for a project with many textX grammars. You might consider making a new tool that extends textX command (with recently introduced extensibility) with the new command that implements this feature.

Do you mean as separate project - or just as part of the core project? If "core project", how do we organize these tools? Should we move export.py into a tools/export.py folder, together with other tools, such as  render_all_grammars.py?

igordejanovic:
> Do you mean as separate project - or just as part of the core project? If "core project", how do we organize these tools? Should we move export.py into a tools/export.py folder, together with other tools, such as render_all_grammars.py?

I'm more for making separate tool. Actually, I would like for all CLI to move out of core textX lib down the road as those are really something needed only during development not in run-time.

goto40:
> > Do you mean as separate project - or just as part of the core project? If "core project", how do we organize these tools? Should we move export.py into a tools/export.py folder, together with other tools, such as render_all_grammars.py?
> 
> I'm more for making separate tool. Actually, I would like for all CLI to move out of core textX lib down the road as those are really something needed only during development not in run-time.

Ok. That means we leave the script in the examples for now? Or do you prefer to not having it there? We could move it to an export tool package once we have defined it...

goto40:
Done. Except the example script. I suggest to leave it there and to move it to the "textx_export" cli-package once we have it?

The changes could be squashed together...

goto40:
Btw.: How do we create new (sub-) projects? E.g. an "textx-export" project. Who is initiating the new repositories on github?

igordejanovic:
@goto40 Thanks! Will do a detailed review soon.
I've created https://github.com/textX/textX-dev-tools project for development tools. We might start to move all CLI/dev stuff there and have the core lib. `textX-dev-tools` would provide `textx` command with all CLI exporters/visualizers/checkers and CLI extensibility to support third-party additions.

It is quite a chunk of work so not sure we should do that for the next 2.0 release. I would like to have that release in the next several weeks as we haven't had one for a long time. We could start to migrate things there but I would like us to wrap up what we have at the moment, to merge all PR that are going on and that should go into the new release. After 2.0 is out we can do the split.

igordejanovic:
> Ok. That means we leave the script in the examples for now?

Yes, I don't mind having it there until we split the tooling part.

goto40:
> It is quite a chunk of work so not sure we should do that for the next 2.0 release. I would like to have that release in the next several weeks as we haven't had one for a long time. We could start to migrate things there but I would like us to wrap up what we have at the moment, to merge all PR that are going on and that should go into the new release. After 2.0 is out we can do the split.

Ok. Let me know what aspects you would prioritize in this PR. I will be happy to make changes/corrections.

FYI: Since the export feature is not easily testable by unittests, this PR may contain bugs (like the ones I found by "inspecting" manually the exported DOT files...). But in my opinion it should be more or less ok. We can still fix bugs in 2.x.x. ...

docs/textx_command.md
```diff
@@ -27,6 +27,21 @@ To get basic help:
...
+    $ textx visualize --help
+    Usage: textx visualize [OPTIONS] META_MODEL_FILE [MODEL_FILE]
+    
+      Generate .dot file(s) from meta-model and optionally model.
+    
+    Options:
+      -i, --ignore-case               case-insensitive parsing.
+      -f, --output-format [dot|plantuml]
+                                      select the output format (plantuml not
+                                      available for model files, yet.
```
igordejanovic, line 41: Missing closing parentheses in the help description.

igordejanovic (approved):
Looks great.  Just a small typo.

Please merge when ready and update CHANGELOG.

BTW, just a though while looking at generated images. I think we should add a new optional flag for `visualize` command that would make rendering of base types optional (e.g. `--base-types`). By default base types would not be rendered in the output. I find that part too distractive and even annoying, especially for a user that already know the base types.

Of course, this is a separate issue. Will create an issue for that.

goto40:
merging

Issue #167, opened 2019-02-16
Reference of create


aranega:
Hi @igordejanovic, 

I was playing with textX and its wonderfu reference system, and I wonder if it is possible to have a rule reference that creates the object first if it does not find it (e.g, like in dot or plantUML). I know that could lead to a mess in the produced model, but I wonder if it currently exists a way of performing such a trick.

Thanks!

igordejanovic:
Hi @aranega.
I believe this could be done by implementing a custom scope provider that would search the model for the given reference and create the object if is none is found. 
Please see [the docs](http://textx.github.io/textX/stable/scoping/).

This part is done by @goto40 so maybe he could give you a hint?

aranega:
@igordejanovic Thanks so much for the pointer, I was sure that I was missing something!

goto40:
I created an example. However, I see some limits. The example will, thus, not work in more complicated scenarios. There, you may need to return a "Postponed" for other lookups for persons, which do not find their reference (since the reference *may* be defined by a later lookup). See #168

goto40:
With this change, we overcome the problem when referencing objects invented by another lookup (in the future) with our postponed mechanism. See example and the docu in the test.

aranega:
Thanks a lot @goto40 . I will play with that today. For my scenario it should be enough. The idea is to bridge plantUML syntax with pyuml2 (UML2 implementation in Python). I'm sure it should be enough! Thanks again!

PR #168, opened 2019-02-16
#167: added example

goto40:


tests/functional/regressions/test_issue167.py
```diff
@@ -0,0 +1,113 @@
...
+
+def person_definer_scope(knows, attr, attr_ref):
+    m = get_model(knows)  # get the model of the currently processed element
+    name = attr_ref.obj_name  # the name of currently looked up element
+    found_persons = list(filter(lambda p: p.name == name, m.persons))
+    if len(found_persons) > 0:
+        return found_persons[0]  # if a person exists, return it
+    else:
+        mm = get_metamodel(m)  # else, create it and store it in the model
+        person = mm['Person']()
```
goto40: Is this ok?
Is it ok to instantiate a model object that way? We could maybe add something in the docs for that...

(also check the next lines, where "name" and "parent" are set...

goto40:
(see comment in #167)

goto40:
Just one unittest. I think it is a very valuable addition motivated from @aranega.
If you have no further comments (e.g., on creating the model instance, see my inline comment), this PR could be squashed...

igordejanovic:
@goto40 Yes, this test is really good. But, I wouldn't put it in regression tests as it is not a bug but demonstrates a feature that was already there but haven't been though of until @aranega asked for. 
I suggest to move this test to core scoping tests and add the part of the docs which shortly explains the use-case and links it to the test (and the issue that motivated it).

tests/functional/regressions/test_issue167.py
igordejanovic: I think this style of instantiation is perfectly fine.

igordejanovic (changes requested):
Please see comments.

goto40:
I do it soon.

goto40:
I moved the test and I wrote some docu. You could have a look. I will be happy to add more aspects and correct things (comments welcome).

docs/scoping.md
```diff
@@ -228,3 +228,43 @@ references to be resolved until all references are resolved or no progress
...
+a reference", you must take into account that these elements
+may have not yet been created. This can be achieved in the same 
+way as handling unresolved references in a scope provider (with the 
+```Postponed``` mechanism). This use case was motivated by 
+[#167](https://github.com/textX/textX/issues/167).
+
+An example of such a meta model is given in 
+[tests/test_model_modification_through_scoping.py](https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_model_modification_through_scoping.py):
+Here you can 
+ * **define** Persons explicitly (existence) and
```
igordejanovic, line 285: After `Here you can` and additional newline is needed to render the bullets list properly.

docs/scoping.md
```diff
@@ -228,3 +228,43 @@ references to be resolved until all references are resolved or no progress
...
+ * **reference** two persons which **know** each other (relationship). 
+ Moreover, **referencing a not existent persons** (all person explicitly defined
```
igordejanovic: `not existent` -> `nonexistent` ?  There are several instances of it.

igordejanovic (approved):
@goto40 Great work! I really like the test example.
There are just some typos I noticed (please, see inline comments).

Please squash merge when ready.

igordejanovic:
BTW, we could introduce the practice of updating the CHANGELOG with each PR (as a part of it). It would make it easier when the time of release comes, and it would make apparent at any time what is merged to the master branch.

goto40:
**Typos fixed**.

Sorry for updating some more:
I added some text between 

 * "Using the scope provider to modify a model" 
 * and "Use case: reference data "defined by references".

Maybe you can have look again before merging... It is some architectural background explaining the relationship between parsing/reference resolution/object processors...

Maybe you could have a quick look at the additional paragraph...

goto40:
I will push the changelog changes

goto40:
I squash and merge after your "ok". Take your time... I appreciate more comments of course.

goto40:
I will merge now. We can reopen a new ticket later. My additional changes were not so relevant.

PR #173, opened 2019-03-03
Fix for issue #155

igordejanovic:
Fixing circular rule references for issue #155.

This PR is based on #159


igordejanovic:
@goto40 I believe that this PR solves #155. It is based on the test from #159 

The problem was that the function `_determine_rule_types` hasn't support circular references in grammar definitions and thus was dependent on the order of the resolution. Now, it is fixed to implement multi-pass algorithm and thus should support arbitrary depth circular references.

I have run the test multiple times on the various Python versions and it looks like the problem has been solved.

@Aluriak @goto40 Thanks much for your help.

If this is indeed a fix for the #155 issue I'll backport the fix to the 1.8.x also and release that bugfix version together with 2.0.0.

textx/lang.py
```diff
@@ -364,9 +366,13 @@ def _add_reffered_classes(rule, inh_by, start=False):
                     else:
                         _add_reffered_classes(rule, cls._tx_inh_by, start=True)
 
-        resolved_classes = set()
-        for cls in metamodel:
-            _determine_rule_type(cls)
+        # Multi-pass rule type resolving to support circular rule references.
+        has_change = [True]
```
goto40, line 372: I believe the use of the array here is a trick to prevent accidental redefinitions of the variable, when assigning a new value. If this is not 100% standard, I suggest to put a small comment.

goto40 (approved):
Your arguments sound plausible. The code you changed is inline with your arguments as far I understand the surrounding algorithms.

I think this part of the TextX code is one of the most demanding parts with respect to deep insider knowledge.

Great work. I appreciate that we are able to understand, and thus maintain the entire code base.

You can delete #159.

textx/lang.py
igordejanovic: Actually, this is a trick for Python 2 to support changing the value of outer scope variable. In Python 3 there is a keyword `nonlocal` which declares that the variable is to be found in the outer scope and the inner functions may actually set this reference. There is no such thing in Python 2. There is a kw `global` but it specifies that the variable is to be found in the global (module-level) scope. But the trick used here is that, although you can't set the variable from the outer scope, you can access/read them. Thus, the reference to the list `has_change` is read inside of the inner function and only the content of the list has been changed.

When we drop support for Python 2 one day, this can be changed to use `nonlocal`.

I've added a short comment in the code.

Issue #171, opened 2019-02-21
Preparing release v2.0.0


igordejanovic:

goto40:
What has to be done for 2.0.0?

igordejanovic:
I guess we should resolve pending PRs (at least what is possible at the moment) and update CHANGELOG with note on all backward incompatible changes.
There is a lot of your work in the pending PRs so I must run through all of those and see what is ready. I will look into it in the following days.

goto40:
Understood. Ok.

zilakmarian:
Will v2.0.0 support python2.7 or are there plans on dropping python2 compatibility?

igordejanovic:
@zilakmarian There are no plans for dropping Python 2 support yet. We will support it as long as there is a support in the tools we are using for development/testing. 

v2.0.0 will be just a start of using semantic versioning where we'll increase major version number with each backward incompatible change. So expect that major versions will increase much faster now, but it should be safe to do upgrades with each minor version change as it shouldn't introduce any backward incompatible change. This way we are better communicating when users should pay attention to the CHANGELOG to see what changes are necessary during upgrade.

PR #174, opened 2019-03-09
Rewrite CHANGELOG to follow keepachangelog format

igordejanovic:
Also updated/added all notable changes after v1.8.0.

refs #171

## Code review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,35 +1,68 @@
 # textX changelog
 
-* Next Release 1.?.?
-   - https://github.com/textX/textX/pull/168
-     - Adding examples and documentation related to scope providers
-       (related to model modification through scope providers).
-   - https://github.com/textX/textX/pull/165
-     - Added metamodel export feature for PlantUML.
-   - https://github.com/textX/textX/pull/134
-     - Fixing bug when using a sequence of matches and rule reference in
-       an abstract rule choice alternative.
-     - Explicitly disallowing referencing more than one rule in an abstract
-       rule choice alternative.
-   - https://github.com/textX/textX/pull/114
-     - python like imports (named import rules, scope providers affected)
-     - OS incompatibility fixes (path separator).
-   - https://github.com/textX/textX/pull/98
-     - Added STRICTFLOAT as buildin type to allow to distinguish ints from
-       floats in NUMBERs. Fixed docu link.
-   - https://github.com/textX/textX/pull/93
-     - Changed attribute name for the metamodel object (from 
-       "metamodel._parser" to "metamodel._parser_blueprint").
-   - https://github.com/textX/textX/pull/120
-     Unicode requirement for (meta)-model strings API parameters made strict.
-     This should prevent common errors with Python 2.x
-     See:
-     https://github.com/textX/textX/issues/105
-     https://github.com/textX/textX/pull/99
-     https://github.com/textX/textX/issues/117
+All _notable_ changes to this project will be documented in this file.
+
+The format is based on _[Keep a Changelog][keepachangelog]_, and this project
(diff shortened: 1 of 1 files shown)


igordejanovic:
@goto40 Please read the new CHANGELOG. Also, a new template for PRs is in place that should help in raising the quality of our code and git log history. The idea is that reviewer should check and tick all checks boxes above while doing the review.

goto40:
I will do it. Good checklist!

CHANGELOG.md
```diff
@@ -1,35 +1,68 @@
...
-     https://github.com/textX/textX/issues/117
-
-
-* 2018-10-06 Release 1.8.0
+All _notable_ changes to this project will be documented in this file.
+
+The format is based on _[Keep a Changelog][keepachangelog]_, and this project
+adheres to _[Semantic Versioning][semver]_.
+
+Everything that is documented in the [official docs][textXDocs] is considered
```
goto40, line 8: In principal this is OK for me. That is how I understand the responsibility of official docs. However, 'everything' is a hard word. Also all referenced unittests? With all details? 

We should at least exclude _underscore properties/functions, except the _tx_* aspects. This gives us the ability to mark implementation details in the future.

goto40: Again, in principal it is OK for me! We should try hard to document all changes! So maybe ignore the last comment.

CHANGELOG.md
```diff
@@ -1,35 +1,68 @@
...
+
+  - CLI migrated to the [click] library ([#162])
+  - Docs improvements ([#146], [#153], [#151]). Thanks simkimsia@GitHub.
+  - `FQN` constuctor param `follow_loaded_models` removed and introduced
+    callback `scope_rediction_logic` to implement arbitrary logic on FQN
+    resolution ([#133], [#114], [#103]) **(BIC)**
+  - Changed attribute name for the meta-model object (from `metamodel._parser`
+    to `metamodel._parser_blueprint`). ([#93]) **(BIC)**
+  - Started using _[Keep a Changelog][keepachangelog]_ ([#?])
+  - Started using _[Semantic Versioning][semver]_ ([#?])
```
goto40: References with '?'

goto40 (changes requested):
Very good! From my side the checklist is a great add-on.

One thing: Some references have '?'. See inline comments. 

No additional critical comments.

CHANGELOG.md
igordejanovic: Yup, we should somehow document the what should be considered a stable API for which change we shall increase the major version number. I guess the correct way is to document stuff that are user-facing and whenever we change those element is a backward incompatible way we do major number increase.

We should also make clear the policy we use to maintain old major versions. I guess we should give enough time for users to migrate to new major versions. Maybe to provide bugfixes to the next to last major version. E.g. when we release 2.0.0 we might provide bugfixes for 1.8.x until we hit 3.0.0. That should give plenty of time for the upgrade but will increase maintenance overhead. Not sure that we have a capacity for that.

CHANGELOG.md
igordejanovic: Good catch! Yeah, I put a question mark with an idea to replace with a proper PR number as soon as I make one, but forgot to update. Will do.

CHANGELOG.md
igordejanovic: Even without support for previous version, users should be safe to declare their dependencies against the major version and receive all backward compatible improvements and bug fixes. So they can stay at previous version as long as they like. When the time is to upgrade to the next major version they should investigate the changelog for all breaking changes and see if they are affected. That should provide (I hope) smooth upgrades.

goto40: OK for me.

goto40 (approved):

Issue #175, opened 2019-03-29
Language registration/discovery


igordejanovic:
I'm starting this discussion as we are in a growing need of organizing multiple
languages/meta-models in a consistent way.

There was [the project](https://github.com/textX/textx-tools) that meant to deal
with the problem of consistent organization/registration and discovery of
languages and compiler/generators. The project has been dormant for a longer
time now and I think it is the time to revive the discussion and see what would
be the proper way to handle this.

What we need is a way to:

-   Register language with its name, file extension and the description
-   Once registered language should be discoverable and meta-model instance should
    be easily retrievable

@goto40 `validate` command from the example is very similar to the current textx
`check` command but done without access to the grammar. With language
registration it would be fairly easy to make `validate` command generic. The
proper language could be selected by the extension or given explicitly to
override the extension.

Something like:

    textx validate --language flow *.flow

In the same way textX grammar could be validated as it is nothing more but a
model written in textX language which should be built-in.

    textx validate flow.tx
    # or
    textx validate --language textx flow.tx

Like it was done in [the old project](https://github.com/textX/textx-tools) we shall have a way to list all the
language with its registered names, extensions and descriptions:

    $ textx list-languages
    textx (tx)    textX grammars
    robot (rbt)   A language for moving robot in a grid
    flow (flow)   A language for description of data processing flow
    ...

Now, compiler/code generator can also be registered in the core. Each generator
should provide the name of the language it is generator for, its name (id), the
description, and the callable that would carry out code generation during
registration.

All registered generators could be listed:

    $ textx list-generators
    flow-python          Generating Python code for flow models
    flow-java            Generating Java code for flow models
    flow-PlantUML        Generating PlantUML visualization for flow models
    robot-java           Generating Java code for Robot models
    robot-django         Generating Django web application for Robot models

Generator can be called on the model(s) for a given language. 

Something like:

    textx generate flow-python <output-folder> *.flow

One nice thing to note is that current `visualize` command is nothing more but a
code generator for textX language and dot/PlantUML output.

@danixeee The additional benefit is that textX-LS integration should be even
simpler as the core language information is provided by the textX itself and
only LS specific stuff should be provided by the current `LanguageTemplate`
inherited class.

This is just some initial brain-dump and surely would require some polishing.

Thoughts?

goto40:
First, I like the ideas you sketched above very much!!

The following points come to my mind:

[...]

PR #177, opened 2019-03-30
WIP: Feature - language registration/discovery

igordejanovic:
This PR is WIP on #175 

@goto40 @danixeee I'm publishing PR to get early feedback as I go along.

## TODO Checklist

- [x] registration API tests (@igordejanovic)
- [x] metamodel `reference` tests (@igordejanovic)
- [x] remove `visualize` command in favor of `generate` (@igordejanovic )
- [x] registration API docs (@igordejanovic)
- [x] meta-model `reference` docs (@igordejanovic)
- [x] `MetaModelProvider` removal and scoping tests update (@goto40, #178)
- [x] Scoping/multi-meta-model docs update (@goto40)
- [x] Allow passing in custom params to generators (@igordejanovic , see [this](https://github.com/textX/textX/issues/175#issuecomment-481322984))

## Code review checklist


danixeee:
@igordejanovic Should you add package `textx.cli` to [setup.py](https://github.com/textX/textX/blob/feature/language-registration/setup.py#L49) ? CLI is not working when I install _textX_ with `pip install .` .

igordejanovic:
@danixeee Yup. Good catch. Thanks.

igordejanovic:
Ok, this branch is getting into shape.

@goto40 I would really appreciate some help here to migrate scoping to the new registration API. The  idea is to replace `MetaModelProvider`. Please see changes in `tests/functional/test_metamodel/` tests and project examples in `tests/functional/registration/projects`. Those project are based on [multimetamodel example](https://github.com/textX/textx-multi-metamodel-example) so the example should be updated accordingly. You can branch your work of this feature branch and make PR against this  branch.

Check how `check` command is now agnostic whether you are checking model or meta-model (each .tx file is a model of textX language/meta-meta-model :) ).  But still you can use command like before, with unregistered language/meta-model from `.tx` file.

All textX commands are registered and treated in the same way. No magical commands. See `setup.py` for what is registered. I still have to migrate `visualize` to be just a plain generator for `dot` and `PlantUML` targets. This way we could move all those textX commands to a separate Python package/project if we decide to. Also, new commands can be added by third-party modules that will be no different from textX provided.

Still have some work to be done but I think this approach will be much cleaner and more consistent.

igordejanovic:
I made scripts : 
- `install-all-dev.sh` - install textX in development mode and install all test projects
- `runtests.sh` - runs all tests, coverage and flake8

These scripts are now called from `.travis.yml`

igordejanovic:
@goto40 One important note: `referenced_metamodels` is removed from APIs (`metamodel_from_str`, `metamodel_from_file`) as it is now handled explicitly in grammars by `reference` keyword (e.g `reference types-dsl as t`).

goto40:
I will have a look tomorrow... Wow. I am looking forward!

danixeee:
@igordejanovic I am writing a generator for VS Code extensions and I noticed that a language id is case-sensitive when I called `generate .../workflow.tx --target vscode` command. I guess that the generator should not be displayed when listing generators or _language_ field in `GeneratorDesc` can be case insensitive.

image

image

P.S. Great work, it's really easy to use and understand! :)

goto40:
Before i dive into details: the responsibility of the meta model provider was to deduce the meta model when importing a model file, based on its file extension.

Do you think this responsibility is still valid? If not, we must find another way to find the meta model for an importuri field...

igordejanovic:
@danixeee Yes, I don't think there is any point in making language and target names case-sensitive. Will change to insensitive matching.

@goto40 registration API now takes over all responsibilities of MetaModelProvider and adds more like generator registration/discovery, declarative registration of languages and generators in setup.py etc. For getting a metamodel from file or pattern use `metamodel_for_file`. See `textx.registration`. The API is exported in `textx` module and should be imported from there.

goto40:
I removed the MetaModelProvider and fixed tests. See "major changes" in #178... we should discuss the changes.

**I really appreciate the new way of registering languages! This is a great addition to textx.**
However, I have no good feeling when clearing all languages (we had the same problem with the - now obsolete - MetaModelProvider)... Should we then at least automatically add some default langues again ("*.tx")?

I think we should add a default metamodel to the importURI scope provider somehow - or **allow to specify the default (or fixed)  metamodel to be used for an importURI statement**... At the moment I have no good idea for this - (the "import rule" must provide that information)... **Or maybe** we need a **metamodel specific metamodelprovider**?

igordejanovic:
> However, I have no good feeling when clearing all languages (we had the same problem with the - now obsolete - MetaModelProvider)... Should we then at least automatically add some default langues again ("*.tx")?

That is perfectly fine. Clearing registration would bring the state of registry to the initial and the first access to the API will re-register declaratively registered languages (using the extension point). Since the textX language is registered as any other language it'll be always there. Check this for example:
```python
clear_language_registrations()
l = language_description('textx')
assert l.name = 'textX'
```

> I think we should add a default metamodel to the importURI scope provider somehow - or allow to specify the default (or fixed) metamodel to be used for an importURI statement... At the moment I have no good idea for this - (the "import rule" must provide that information)... Or maybe we need a metamodel specific metamodelprovider?

I'm repeating here a comment from #178 to be in context. I'm thinking about providing an optional syntax for `importURI` that could specify the language. For example:
```
import 'flow-dsl:/some/path/myflow.flow'
``` 
If the `<language>:` prefix exists than `metamodel_for_language` is called else `metamodel_for_file` is called to get the meta-model.

Another addition (to be consistent with the current `reference` keyword in textX language) is to enable:
```
import 'flow-dsl:/some/path/myflow.flow' as mf
```
And than we could use qualified names to `mf.SomeObject`. I think we have discussed this somewhere. Perhaps we could make a new TEP on this to collect information scattered around in one place?

igordejanovic:
@goto40 I've added a TODO checklist at the top of the stuff need to be done before this is ready for merge.

igordejanovic:
> Or maybe we need a metamodel specific metamodelprovider?

That is something to think about. We might use now explicit meta-model references through `reference` keyword to deduce needed meta-model even if it is not explicitly specified in importURI. That is probably something for another PR, maybe post 2.x.

goto40:
Good work!

igordejanovic:
@goto40  Thanks :)

goto40:
I would like to discuss, if we could not use the python package system instead of our custom language registration module?

Python is good in finding packages. I would like to reuse this functionality for finding languages.

What do both of you think about this? It should not be a big deal to change this at the current state?

goto40:
For the user it just means to use the full package name or a unique substring of it (at the command line) to identify a language.

For the meta model designer it means not to have to think about a good language name, but about a good package name only.

igordejanovic:
Not sure that I understand but if you propose a dynamic load using a FQN of a Python package then I'm not convinced that it would be a good way to go. Here are a couple of reasons:
- If you are concerned about name clashing think that our languages would share the same namespace with all the Python packages. I generally don't think that we need some complex FQN naming scheme. If textX become so popular that we have so many languages around and the name clashing becomes a real problem... well that's the problem I would like to have :smile: 
- Python dynamic import has considerably changed from version to version so we would have to support all variants used in Python versions we support
- You would still need to register your module either declaratively or imperatively in order to be able to list it (`textx list-languages`) so I'm not really able to see any benefit in comparison to the current approach

goto40:
OK. It was just an idea.

igordejanovic:
Sure, no problem. I like to discuss new ideas :)

goto40:
Maybe we could allow to pass an optional list of LanguageDesc when creating a meta model.

This list could override global registration in case of conflicts... We could (or I could the next days) encapsulate the registration.py logic in an object. This object would then exist one time globally and could be created locally for individual meta models in an multi meta model context.

The global registration api is untouched. 

The local lookup/registration object could be accessible via _tx_attribute for, e.g., editors.

Local registration could be automatically forwarded to global registration, if no conflicts exist.

igordejanovic:
What problem would that solve?

igordejanovic:
If that would solve only the potential conflicts then I suggest to push that back to after 2.0 as it wont be a backward incompatible change so we could do that for 2.1. I would like to focus now on getting 2.0 out. I was about to do the release but figured out that this registration/discovery is important to be done ASAP to unblock @danixeee work on textX-LS.

If @danixeee can continue his work with this we can focus on updating tests and docs and releasing 2.0. After that we'll have more time to make a proper decisions and improvements. I also have some ideas on how to extend the possibilities of newly introduced `reference` keyword.

goto40:
The use case I had in mind is the one you mentioned : a language l1 uses a language l2. If I install 'another' language l2 I break my language l1.

Local registration solves this problem. I can make a pull request after 2.0. Or we can discuss...

igordejanovic:
Ok. I suggest to move the discussion to a new issue and discuss the solution before the implementation.

danixeee:
> I was about to do the release but figured out that this registration/discovery is important to be done ASAP to unblock @danixeee work on textX-LS.

Yeah, I stopped with `textX-LS` for now, because this implementation allows us to create separate projects (generators) which will be used by language server, so I am just playing with generators for now. Once we address passing [custom arguments](https://github.com/textX/textX/issues/175#issuecomment-481322984) to generator, I will continue my work.

igordejanovic:
@danixeee Generators now accept custom arguments. See the last commit.

danixeee:
@igordejanovic I have some problems getting custom arguments from CLI, can you take a look when you have chance?

Project: https://github.com/danixeee/textx-gen-coloring
Install: `pip install .`
Command: In Readme

Custom arguments dictionary is always empty [here](https://github.com/danixeee/textx-gen-coloring/blob/master/textx_gen_coloring/__init__.py#L30).

igordejanovic:
@danixeee It was a bug in custom arguments collecting. Fixed.

danixeee:
I have implemented a basic [textmate generator](https://github.com/danixeee/textx-gen-coloring) for syntax highlighting. It works with the latest changes from this branch.

igordejanovic:
@danixeee Cool! Any docs on coloring DSL? How can we try the generator? I guess the generated highlighting file should be copied somewhere in VS Code installation or workspace or...?

I suggest to put a note in the README that this branch is needed if someone wants to test it before this PR is merged.

danixeee:
@igordejanovic I have implemented just default syntax coloring generator, which uses targeting language grammar and produces json file. It collects just strings (keywords) from a grammar.
That generator is called from [textx-gen-vscode](https://github.com/danixeee/textx-gen-vscode). For now, output is `.vsix` file (extension which can be installed). I will add `--install` flag, if we want to install the extension and delete `.vsix` file afterwards.

VS Code generator will be called from `textX-LS` in a subprocess, and after successful installation, if VS Code is open, `Reload` button will be displayed and user will have to reload the editor (there is no workaround for it AFAIK).

I will keep those generators in my repository until this PR is merged to master, then I will add test/docs and move them under `textX` organization.

Do you plan to include work on this PR in the next release?

igordejanovic:
@danixeee Sounds great.

> Do you plan to include work on this PR in the next release?

Yes, I'm planning to release 2.0.0 as soon as this PR is fully finished and merged.

igordejanovic:
@goto40 Regarding my previous comment, I see that you have ongoing work on `scope_provider_interactive_demo2`. Is it something you plan to merge before v2?

goto40:
I want to add more unittests and refine things. I need some more days. Possibly we will need also to discuss things with @danixeee. 

However, the critical part was already merged with #186. My ongoing work should not break anything. So we can go ahead with textX 2.0.0...

igordejanovic:
@goto40 I think I have finished this PR mostly. What remains is the update of scoping/multi-metamodel docs. I did a rework of introduction and first use-case in `multimetamodel.md`. 

Could you take a look when find some time and see if what I did is ok and could you update the rest of the docs?

goto40:
I am happy to do it.

textx/registration.py
```diff
@@ -0,0 +1,312 @@
...
+    Args:
+        language_desc_or_name (LanguageDesc or str): If LanguageDesc is given
+            other parameters are not used.
+        For other parameters see `LanguageDesc`.
+    """
+    global languages
+    if languages is None:
+        language_descriptions()
+
+    if type(language_desc_or_name) is not LanguageDesc:
```
danixeee: @igordejanovic Can you change this to check if `language_desc_or_name` is instance of `LanguageDesc`?
The reason is that we will need additional fields/methods for language description for `textX-LS`, and I thought it would be a good idea to avoid replacing similar logic in `textX-LS-core` and to allow `textX` to handle subclasses of `LanguageDesc`.

textx/registration.py
```diff
@@ -0,0 +1,312 @@
...
+        self.description = description
+        self.generator = generator
+
+
+metamodels = {}
+languages = None
+generators = None
+
+
+def language_descriptions():
```
danixeee, line 57: A flag parameter like `reload_anyway` will be useful, as well for `generator_descriptions`.
I had to call `pkg_resources.WorkingSet().iter_entry_points(group='textx_languages')` in order to discover newly installed languages in already running process.

danixeee: I see now that you have already added functions to clear languages and generators, sorry.

textx/registration.py
igordejanovic: Sure. It sounds like a good idea.

textx/registration.py
igordejanovic: Yes. When the registration is cleared declaratively registered registrations gets reloaded with the next call to the API.

danixeee: I think we still need to change to:

_pkg_resources`.WorkingSet()`.iter_entry_points(group='textx_languages')_

to reinitialize _entries_ field of working set, because `pip install ...` adds a new path on `sys.path`. Also, in `generator_descriptions` function.

https://github.com/pypa/setuptools/blob/8d4e4bc93e319275458dcc7381ec83d447eca770/pkg_resources/__init__.py#L564

igordejanovic: You are right. Fixed. Thanks.

textx/registration.py
igordejanovic: Done, for both languages and generators.

igordejanovic:
I think we are now in a good shape with this PR.
@goto40 @danixeee Is there anything on your end?

If everything looks ok I'll be working on merging this branch and releasing v2 tomorrow.

goto40:
No more comments from my side. Looking forward to see v2!

danixeee:
Looks good to me as well. Check this out: https://github.com/textX/textX-LS/issues/13#issuecomment-493467118. :)

danixeee:
@igordejanovic Can we add project name to `LanguageDesc` and `GeneratorDesc` objects during entry point iteration?

igordejanovic:
@danixeee Yes, I think that would indeed be a valuable information. Done on `master` in commit  9d4fa74b7f77479f39774070c7f526d968078d41

PR #178, opened 2019-04-02
Feature/language registration meta model provider replacement

goto40:
A quickfix proposal for #177: "removed MetaModelProvider", fixed some tests.

## Major changes
 * removed MetaModelProvider
 * allow - in addition to a metamodel factory function - also to register a metamodel itself. This should
allow to register a concrete instance as well... **comments?**
 * Fixed importUri metamodel retrieving: allow old behavior to use the current metamodel as fallback: **TBC**: can we have a check function for the existence of a language (instead of try/except, as I implemented it). -- see: ```metamodel_for_file_or_default_metamodel``` in textx/scoping/_ _ init _ _ .py

## Minor changes 
 * fixed some tests regarding model registration

## Misc
 * Note for the commit messages: we can squash everything together, after we reviewed and adapted the code...
 * Tests are OK now, but coverage is below 80%

## Code review checklist

### DIFF
--- a/tests/functional/regressions/test_issue140.py
+++ b/tests/functional/regressions/test_issue140.py
@@ -1,7 +1,8 @@
 from __future__ import unicode_literals
 import os
 import os.path
-from textx import metamodel_from_file
+from textx import metamodel_from_file, LanguageDesc,\
+    register_language, clear_language_registrations
 import textx.scoping.providers as scoping_providers
 import textx.scoping as scoping
 
@@ -36,9 +37,21 @@ def test_multi_metamodel_obj_proc():
     mm_A.register_obj_processors({"C1": proc})
     mm_B.register_obj_processors({"C1": proc})
 
-    scoping.MetaModelProvider.clear()
-    scoping.MetaModelProvider.add_metamodel("*.a", mm_A)
-    scoping.MetaModelProvider.add_metamodel("*.b", mm_B)
+    a_dsl = LanguageDesc(
+        name='a-dsl',
+        pattern='*.a',
+        description='Test Lang A',
+        metamodel=mm_A)
+
+    b_dsl = LanguageDesc(
+        name='b-dsl',
+        pattern='*.b',
+        description='Test Lang B',
+        metamodel=mm_B)
+
+    clear_language_registrations()
+    register_language(a_dsl)
+    register_language(b_dsl)
 
     mm_B.model_from_file(os.path.join(
         os.path.dirname(__file__),
--- a/tests/functional/regressions/test_issue140.py
+++ b/tests/functional/regressions/test_issue140.py
@@ -37,21 +37,17 @@ def test_multi_metamodel_obj_proc():
     mm_A.register_obj_processors({"C1": proc})
     mm_B.register_obj_processors({"C1": proc})
 
-    a_dsl = LanguageDesc(
-        name='a-dsl',
+    clear_language_registrations()
+    register_language(
+        'a-dsl',
         pattern='*.a',
         description='Test Lang A',
         metamodel=mm_A)
-
-    b_dsl = LanguageDesc(
-        name='b-dsl',
+    register_language(
+        'b-dsl',
         pattern='*.b',
         description='Test Lang B',
         metamodel=mm_B)
-
-    clear_language_registrations()
-    register_language(a_dsl)
-    register_language(b_dsl)
 
     mm_B.model_from_file(os.path.join(
         os.path.dirname(__file__),
--- a/tests/functional/test_scoping/test_exception_from_included_model.py
+++ b/tests/functional/test_scoping/test_exception_from_included_model.py
@@ -52,28 +52,22 @@ def test_exception_from_included_model():
         import_lookup_provider, join(this_folder,
                                      "metamodel_provider3", "C.tx"))
 
-    a_dsl = LanguageDesc(
-        name='a-dsl',
+    clear_language_registrations()
+    register_language(
+        'a-dsl',
         pattern='*.a',
         description='Test Lang A',
         metamodel=a_mm)
-
-    b_dsl = LanguageDesc(
-        name='b-dsl',
+    register_language(
+        'b-dsl',
         pattern='*.b',
         description='Test Lang B',
         metamodel=b_mm)
-
-    c_dsl = LanguageDesc(
-        name='c-dsl',
+    register_language(
+        'c-dsl',
         pattern='*.c',
         description='Test Lang C',
         metamodel=c_mm)
-
-    clear_language_registrations()
-    register_language(a_dsl)
-    register_language(b_dsl)
-    register_language(c_dsl)
 
     #################################
     # MODEL PARSING / TEST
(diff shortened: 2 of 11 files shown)


textx/registration.py
```diff
@@ -193,8 +199,8 @@ def languages_for_file(file_name_or_pattern):
     """
     file_languages = []
     for language in language_descriptions().values():
-        if file_name_or_pattern == language.pattern \
-           or fnmatch.fnmatch(file_name_or_pattern, language.pattern):
+        pmatch = fnmatch.fnmatch(file_name_or_pattern, language.pattern)
+        if file_name_or_pattern == language.pattern or pmatch:
```
igordejanovic: The old version was better as the `or` will short-circuit and won't call `fnmatch` if the first condition is satisfied.

textx/scoping/__init__.py
```diff
@@ -11,58 +11,15 @@
...
-                "unexpected: no meta model found for {}".format(filename))
+def metamodel_for_file_or_default_metamodel(filename, the_metamodel):
+    from textx import metamodel_for_file
+    try:
+        the_metamodel = metamodel_for_file(filename)
+        # TODO, I would prefer to query if the language was found...
+    except Exception:
+        if the_metamodel is None:  # no metamodel defined...
+            raise
+    return the_metamodel
```
igordejanovic: It is actually a preferred way to code in a style `Ask for forgiveness, not permission`.

The body of this function should be:
```python
try:
    return metamodel_from_file(filename)
except TextXRegistrationError:
    return the_metamodel
```

tests/functional/test_scoping/test_exception_from_included_model.py
```diff
@@ -52,10 +52,51 @@ def my_processor(m):
...
+    c_dsl = LanguageDesc(
+        name='c-dsl',
+        pattern='*.c',
+        description='Test Lang C',
+        metamodel=c_mm)
+
+    clear_language_registrations()
+    register_language(a_dsl)
+    register_language(b_dsl)
+    register_language(c_dsl)
```
igordejanovic: Definition and registration of these 3 languages repeats twice.

tests/functional/regressions/test_issue140.py
```diff
@@ -36,9 +37,21 @@ def proc(a):
...
+
+    b_dsl = LanguageDesc(
+        name='b-dsl',
+        pattern='*.b',
+        description='Test Lang B',
+        metamodel=mm_B)
+
+    clear_language_registrations()
+    register_language(a_dsl)
+    register_language(b_dsl)
```
igordejanovic: I'm adding a way to call `register_language` with direct parameters. This allows for shorter form:
```python
register_language('a-dsl', '*.a', metamodel=mm_A)
```

igordejanovic (changes requested):
Looks good. See inline comments.

igordejanovic:
> Fixed importUri metamodel retrieving: allow old behavior to use the current metamodel as fallback

Just a thought. I'm thinking about providing an optional syntax for `importURI` that could specify the language. For example:
```
import 'flow-dsl:/some/path/myflow.flow'
``` 
If the `<language>:` prefix exists than `metamodel_for_language` is called else `metamodel_for_file` is called to get the meta-model.

igordejanovic:
>   Note for the commit messages: we can squash everything together, after we reviewed and adapted the code...

OK, or you can reword commit messages to follow the guideline. 

> Tests are OK now, but coverage is below 80%

That's perfectly fine as the registration tests are TBD on the base branch.

textx/registration.py
goto40: done

textx/scoping/__init__.py
goto40: ok. I follow the style, that exceptions should be exceptional in many cases; e.g., I ask "if exists(file)" instead of handling file not found exception. But this is ok here.

Code adapted.

tests/functional/test_scoping/test_exception_from_included_model.py
goto40: Copy paste... Thank you!

goto40:
* [x] ```TODO: register_language('a-dsl', '*.a', metamodel=mm_A)``` after rebase

tests/functional/regressions/test_issue140.py
goto40: Next step, after rebasing....

goto40:
I though a lot about the importURI / extension problem: 
 * **Use Case**: a language (e.g. cpp-laguage) allows to include another language (e.g. a h-language). 
 * **Desired behavior**: Then the import statement should be configured to use the h-language for *.h files (independently of other installed languages in the system, which may have the same extension).
 * **Proposal**: In addition to the current behavior of the ImportURI-Provider, I suggest to allow to specify an optional map of maps (for the provider):
    ``` { "Rule" --> { "Pattern" --> metamodel }} ```
    *  Rule can a rule name or "*"
    * Pattern is a file pattern (e.g. "*.h")
    * **Behavior**: if an entry is specified for a given Rule, all files incuded by this rule MUST match one Pattern (or yield an error).

**I could easily add the code with this behavior in this PR** (a few lines + some tests). Can easily be reverted! What do you think? Should we wait until the next PR, or should I make the change?

goto40:
I do not think a modeller (not meta-modeller) should specify the language (other than by the file extension).

tests/functional/regressions/test_issue140.py
goto40: Done.

I have a question concerning rebasing:
I did ``` git rebase --onto origin/feature/language-registration ```: is this the way to go? I got a merge (which I did not understand completely)...?

igordejanovic: Since you have branched from `feature/language-registration` you don't have to use `--onto`. `--onto` is used only if you are rebasing on some other branch. See `man git-rebase` for some nice examples.

All commits from your branch are applied on the top of the base branch one by one. It's like you just did those changes/commits manually. Some of them might be in conflict with the changes in the base branch in which case you should resolve the conflicts (see `git status` and search for conflict markers in files). Once resolved you continue the rebase with `git rebase --continue`). At any time  you can abort the rebase and return to the previous state with `git rebase --abort`). If you run rebase in interactive mode you get a chance to squash, reword, split, augment etc. each commit by giving a special commands in the editor. That is a nice tool when you are need to tidy up the branch.

Tip: git keeps track of all branch positions (branches are like pointers) so if you understand this mechanism you can dissect and change your branch history with confidence as all commits you have made are there until garbage collecting kicks in when all dangling commits (without pointers -- refs: branches, tags pointing to them) are removed. To see this log of where your branch pointer was in the past see `git reflog`. And `git reset` can be used to move your branch pointer to any point in history. Of course, these operations should be used rarely and never on the published branch (e.g. never on `master`) as the published history should be immutable.

igordejanovic:
> In addition to the current behavior of the ImportURI-Provider, I suggest to allow to specify an optional map of maps (for the provider)

I would rather think of a way to specify that in the grammar. I have some rough ideas but lets move that discussion to another PR/issue.

> I do not think a modeller (not meta-modeller) should specify the language (other than by the file extension).

Yes, you are right. It should be a responsibility of language designer.

igordejanovic (approved):

tests/functional/regressions/test_issue140.py
igordejanovic: Strange enough. I just investigated the branch history and it looks like you have do a regular merge of `feature/language-registration` to your branch. Never mind ;)

Issue #96, opened 2018-11-03
object_processors do not run on rule


Geniedesalpages:
I have simplified my grammar to reproduce a bug in object processors:

```python
from textx import metamodel_from_str
grammar = """
Command:
    name=ID parameters=ParameterList
;
    
ParameterList:
    '(' parameters*=Parameter ')'
;
Parameter:
     ExplicitFloat
;
ExplicitFloat:
    /-?\d+\.\d+/
;
"""

code = """
mycommand(0.1 1.1)
"""
mm = metamodel_from_str(grammar)
mm.register_obj_processors({'ExplicitFloat': lambda x: float(x)})

mymodel = mm.model_from_str(code)

first_param = mymodel.parameters.parameters[0]
assert isinstance(first_param, float), f"first parameter is of type {type(first_param)} instead of float"
```

The ExplicitFloat is not changed to a float and remains str. If the ExplicitFloat appears "higher" in the syntax tree of the grammar, then it occasionally translates it. Am I doing something wrong, or is there a bug in the object processors?

goto40:
Thank you for your investigation. I think you need to assign the explicit float object to an attribute, e.g., value=ExplicitFloat.

With this change, your program works for me:

 * Grammar change: Parameter: value=ExplicitFloat
 * assert change: assert isinstance(first_param.value, float), f"first parameter is of type {type(first_param)} instead of float"

Note, that we had some issues with object processors. You should update in any case to textX 1.8

goto40:
Note: The rule Parameter is a match rule in my eyes, as ExplicitFloat is... Thus, I assume, the nested match rule ExplicitFloat is not instantiated... 

A related unit test which may help as a workaround is https://github.com/igordejanovic/textX/blob/master/tests/functional/examples/test_modeling_float_int_variables.py.

goto40:
We need to clarify how object processors apply to match rules, especially for nested match rules.

Geniedesalpages:
Thank you for your quick reply!

By changing it to have value=ExplicitFloat in the grammar, i managed to run the above example. In my original grammar (which is more nested and has more intermediate steps) this approach works in some cases but not in others, I couldn't pin down the error as of yet (I am using 1.8.0 for reference).

Is the object processor not applied right after the instantiation of the Rule, independent of where it is in the AST?

My `Parameter` rule in the full grammar actually looks like this:

```
Parameter:
[...]

PR #183, opened 2019-04-27
Fix #182 Better handling of match rule object processors

### LINKED ISSUE #182: Object processors on terminal match rules not called
Issue found in [SO post](https://stackoverflow.com/questions/55862851/how-to-write-a-textx-grammar-rule-to-detect-standard-datatypes-without-modifying/55864879)

To reproduce:
```python
from textx import metamodel_from_str

grammar = '''
Model: commands*=Command;
Command: MyNumber;
MyNumber: MyFloat | INT;
MyFloat: /[+-]?(((\d+\.(\d*)?|\.\d+)([eE][+-]?\d+)?)|((\d+)([eE][+-]?\d+)))(?<=[\w\.])(?![\w\.])/;
'''

mm.register_obj_processors({'MyFloat': lambda x: float(x)})

model = mm.model_from_str('''
10
10.5
''')
assert type(model.commands[0]) is int
assert type(model.commands[1]) is float

```

igordejanovic:
This PR reworks calling of match rules object processors and fix #182. Now, rules are called during type conversion which should simplify thing and make overriding of default type more consistent.

This seems to fix various issues around calling object processors on match rules like the one reported on #96 

## Code review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ please take a look at related PRs and issues and see if the change affects you.
      
 ### Fixed
 
+  - Calling of match rule object processors ([#183], [#182], [#96])
   - Circular rule references in grammars ([#173], [#159], [#155])
   - Assertion error while calling object processors with multi meta models
     (extended grammars) and custom types ([#141], [#140])
@@ -374,6 +375,8 @@ please take a look at related PRs and issues and see if the change affects you.
   - Export to dot.
 
 
+[#183]: https://github.com/textX/textX/pull/183
+[#182]: https://github.com/textX/textX/issues/182
 [#174]: https://github.com/textX/textX/pull/174
 [#173]: https://github.com/textX/textX/pull/173
 [#168]: https://github.com/textX/textX/pull/168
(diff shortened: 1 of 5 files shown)


goto40:
I have a question. Is the model completely constructed and all references resolved before calling object processors (as it was the case before)?

See: https://github.com/textX/textX/blob/master/docs/scoping.md#using-the-scope-provider-to-modify-a-model

As far as I see your changes, it looks OK...

igordejanovic:
@goto40 Thanks for the review.

With this change object processors that operate on match rules are called bottom up during model construction. Those object processors should generally be a very simple, usually just a type conversions so they don't need a fully constructed model.
I moved this part to the parse tree processing because I think it is the right time to do so. Trying to do so on the constructed model is very complex, or in some situation impossible. E.g. with nested match rules, you end-up only with a top level result so all possible calls to object processors in the hierarchy bellow get lost. This looks confusing to the users, and I now agree (see #96). Also, the override of built-in types or implementation of custom type was buggy (see #182).

Other object processors (i.e. those operating on common rules) are called as before on a constructed model, again in bottom up fashion.

goto40:
Very good. Maybe add a note in the referenced doc section above.

igordejanovic:
Sure, I agree. Would you like to do it?

goto40:
OK.

PR #186, opened 2019-04-28
cleanup localscope implementtation (`RelativeName` and `ExtRelativeName`)

goto40:
## Code review checklist

 * **Goal**: "Refactor" local scope providers to make implementation and API clearer.
 * **Problem:** `RelativeName` and `ExtRelativeName` are difficult to explain.
 * **Solution:** refactored both scope providers and some internal functions in textx.scoping.tools.py. 
   * Made implementation more reabale:
       * textx.scoping.providers.RelativeName.get_reference_propositions
       * textx.scoping.providers.ExtRelativeName.get_reference_propositions (here, we have much clarity gained!)
   * Especially one accidentally included feature made the code very difficult to understand: When traversing the model with a "model-path", lists were interpreted as list of named-objects. This is not needed by all examples given.
   *  Thus, it is not a 100% refactoring, since we do not allow to have lists within
a path to a desired object (but this was never used in any example). It is
not very probable that this breaks existing code (the old behavior can reactivated with the parameter 'follow_named_element_in_lists' in tools.py with minor differences for postponed references).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ please take a look at related PRs and issues and see if the change affects you.
   - Made scope provider implementation of `RelativeName` and `ExtRelativeName`
     more readable ([#186]). Minor functional changes, not very probable to have
     any impact (only affects model-paths containing a list not at the end of the
-    path; see [#186]).
+    path; see [#186]). Possible **(BIC)**.
   - CLI migrated to the [click] library ([#162])
   - Docs improvements ([#146], [#153], [#151]). Thanks simkimsia@GitHub.
   - `FQN` constuctor param `follow_loaded_models` removed and introduced
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ please take a look at related PRs and issues and see if the change affects you.
 
 ### Changed
 
+  - Made scope provider implementation of `RelativeName` and `ExtRelativeName`
+    more readable ([#186]). Minor functional changes, not very probable to have
+    any impact (only affects model-paths containing a list not at the end of the
+    path; see [#186]).
   - CLI migrated to the [click] library ([#162])
   - Docs improvements ([#146], [#153], [#151]). Thanks simkimsia@GitHub.
   - `FQN` constuctor param `follow_loaded_models` removed and introduced
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -90,32 +90,32 @@ We provide some standard scope providers:
    name loopup**.
    Example: see [tests/test_scoping/test_full_qualified_name.py](https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_full_qualified_name.py).
    
-   A central feature of this scope provider is, that it **traverses the model
-   tree and searches for a matching sequence of named objects** (objects with
-   an attribute name matching parts of the full qualified name separated by 
-   dots). You can also provide a **callback** (`scope_redirection_logic`) to specify
-   that certain named objects are not searched recursively, but are replaced
-   by a list of objects instead, which are searched in place of the current object.
-   With this feature you can create, e.g., **namespace/package aliases** in your 
-   language. You can also activate a **python like module import behavior** 
-   for your language (with `textx.scoping.providers.FQNImportURI`), which is based 
-   on this callback.
-   Example: see [tests/functional/regressions/test_issue103_python_like_import.py](https://github.com/textX/textX/blob/master/tests/functional/regressions/test_issue103_python_like_import.py).
-
-        package p1 {
-            package p2 {
-                class a {};
+       A central feature of this scope provider is, that it **traverses the model
+       tree and searches for a matching sequence of named objects** (objects with
+       an attribute name matching parts of the full qualified name separated by 
+       dots). You can also provide a **callback** (`scope_redirection_logic`) to specify
+       that certain named objects are not searched recursively, but are replaced
+       by a list of objects instead, which are searched in place of the current object.
+       With this feature you can create, e.g., **namespace/package aliases** in your 
+       language. You can also activate a **python like module import behavior** 
+       for your language (with `textx.scoping.providers.FQNImportURI`), which is based 
+       on this callback.
+       Example: see [tests/functional/regressions/test_issue103_python_like_import.py](https://github.com/textX/textX/blob/master/tests/functional/regressions/test_issue103_python_like_import.py).
+    
+            package p1 {
+                package p2 {
+                    class a {};
+                }
             }
-        }
-        using p1.p2 as main
-        var x = new p1.p2.a()
-        var y = new main.a()
-        
-   Note: Except in the context of the scope_redirection_logic (see above), the FQN does
-   not take Postponed (unresolved) references into account. The reason is that
-   this would create a much more complex decision logic to decide which reference
-   needs to be resolved first. The purpose of the FQN is to identify direct instances
-   of model objects, and no references.
+            using p1.p2 as main
+            var x = new p1.p2.a()
+            var y = new main.a()
+            
+       Note: Except in the context of the scope_redirection_logic (see above), the FQN does
+       not take Postponed (unresolved) references into account. The reason is that
+       this would create a much more complex decision logic to decide which reference
+       needs to be resolved first. The purpose of the FQN is to identify direct instances
+       of model objects, and no references.
    
  * `textx.scoping.providers.ImportURI`: This a provider which **allows to load
    additional modules** for lookup.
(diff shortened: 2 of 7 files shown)


goto40:
Note: I extracted this refactoring from ongoing experiments in https://github.com/textX/textX/tree/feature/scope_provider_interactive_demo. This other branch could lead to a proposal how to generate reference name prorpositions (for an editor).

docs/scoping.md
```diff
@@ -162,7 +162,12 @@ We provide some standard scope providers:
  * `textx.scoping.providers.ExtRelativeName`: The same as `RelativeName` **allowing
    to model inheritance or chained lookups**.
    Example: see [tests/test_scoping/test_local_scope.py](https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_local_scope.py).
-
+    - `RelativeName`: The scope is defined by a model-path from the current object to (typically) 
+        a list containing named objects.
+    - `ExtRelativeName`: The scope is defined by a model-path from the current object and *connected objects*
+        to (typically) a list containing named objects (similar to `RelativeName`). 
+        The *connected objects* are defined again by a model-path from one object 
+        to another named object or a list of named objects.
```
igordejanovic: Should these `RelativeName` and `ExtRelativeName` explanations be merged in the proper sections defined up?

BTW, this section on scope providers is not rendered correctly in the browser. You can test the docs locally with `mkdocs serve`.

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ please take a look at related PRs and issues and see if the change affects you.
 
 ### Changed
 
+  - Made scope provider implementation of `RelativeName` and `ExtRelativeName`
+    more readable ([#186]). Minor functional changes, not very probable to have
+    any impact (only affects model-paths containing a list not at the end of the
+    path; see [#186]).
```
igordejanovic: If there is a possibility to have any backward incompatibility a changelog entry should be marked with `Possible **(BIC)**`.

tests/functional/examples/test_hierarchical_data_structures_referencing_attributes.py
```diff
@@ -46,33 +46,43 @@ def test_referencing_attributes():
...
-                None, reference.refs[index - 1],
-                "valref.type.vals.{}".format(attr_ref.obj_name),
-                attr_ref.cls)
+            if reference.refs[index - 1].valref is None:
+                return Postponed()
+            if reference.refs[index - 1].valref.type is None:
+                return Postponed()
+            x = get_named_obj_in_list(
+                reference.refs[index - 1].valref.type.vals,
+                attr_ref.obj_name)
```
igordejanovic: This whole if/else could be simplified:
```python
base = reference.instance if index == 0 else reference.refs[index - 1]
if base is None or base.type is None:
    return Postponed()
x = get_named_obj_in_list(base.type.vals, attr_ref.obj_name)
```

textx/scoping/providers.py
```diff
@@ -547,22 +547,41 @@ def __init__(self, path_to_container_object):
...
-            line, col = get_parser(obj).pos_to_linecol(obj_ref.position)
-            raise TextXSemanticError('{}'.format(str(e)), line=line, col=col,
-                                     filename=get_model(obj)._tx_filename)
+        from textx import textx_isinstance
+        obj_list = get_referenced_object(obj, self.path_to_container_object)
+        if type(obj_list) is Postponed:
+            self.postponed_counter += 1
+            return obj_list
+        assert obj_list is not None
+        assert isinstance(obj_list, list)
```
igordejanovic: `get_referenced_object` - the name (and the docstring) doesn't imply that the return value is a list.

textx/scoping/providers.py
```diff
@@ -547,22 +547,41 @@ def __init__(self, path_to_container_object):
...
-            from textx.scoping.tools import get_parser
-            line, col = get_parser(obj).pos_to_linecol(obj_ref.position)
-            raise TextXSemanticError('{}'.format(str(e)), line=line, col=col,
-                                     filename=get_model(obj)._tx_filename)
+        from textx import textx_isinstance
+        obj_list = get_referenced_object(obj, self.path_to_container_object)
+        if type(obj_list) is Postponed:
+            self.postponed_counter += 1
+            return obj_list
+        assert obj_list is not None
```
igordejanovic: This assertion could be made at `get_referenced_object` or removed altogether. I don't think it provides some value.

textx/scoping/providers.py
```diff
@@ -547,22 +547,41 @@ def __init__(self, path_to_container_object):
...
+        assert isinstance(obj_list, list)
+        for obj in obj_list:
+            assert hasattr(obj, "name")
```
igordejanovic: These assertion in the loop are also superfluous as it is check in `filter` which follow (i.e. `AttributeError` will be raised if there is an object without `name`).

textx/scoping/providers.py
```diff
@@ -581,32 +600,51 @@ def __init__(self, path_to_definition_object, path_to_target,
...
+            def_obj, self.path_to_extension)
+        # for all containing classes, collect all
+        # objects to be looked up (e.g. methods)
+        obj_list = []
+        for def_obj in def_objs:
+            if type(def_obj) is Postponed:
+                self.postponed_counter += 1
+                return def_obj
+
+            tmp_list = get_referenced_object(def_obj, self.path_to_target)
```
igordejanovic: Please see the comment on `get_referenced_object` above.

textx/scoping/providers.py
```diff
@@ -581,32 +600,51 @@ def __init__(self, path_to_definition_object, path_to_target,
...
+            assert tmp_list is not None
+            assert isinstance(tmp_list, list)
+            for obj in tmp_list:
+                assert hasattr(obj, "name")
```
igordejanovic: Same comments for assertions as above.

textx/scoping/tools.py
```diff
@@ -130,96 +130,80 @@ def get_recursive_parent_with_typename(obj, desired_parent_typename):
...
     from textx.scoping import Postponed
-    assert prev_obj or not type(obj) is list
     names = dot_separated_name.split(".")
     match = re.match(r'parent\((\w+)\)', names[0])
-    if match:
+
+    if obj is None:
+        return None
+    elif type(obj) is Postponed:
+        return obj
```
igordejanovic, line 187: Previous 4 lines could be written:
```python
if obj is None or type(obj) is Postponed:
    return obj
```

textx/scoping/tools.py
```diff
@@ -130,96 +130,80 @@ def get_recursive_parent_with_typename(obj, desired_parent_typename):
...
+    elif type(obj) is list:
+        if follow_named_element_in_lists:
+            next_obj = get_named_obj_in_list(obj, names[0])
+        else:
+            raise Exception(
+                "unexpected: got list in path for get_referenced_object")
```
igordejanovic, line 194: I think it is always better to use `TextXError` (or some subclass of it) for our error so that user of our lib could catch exception at finer granularity.

igordejanovic (changes requested):
Looks good. I like the addition of `get_reference_propositions` as it could be helpful for further tooling development (/cc @danixeee).

There are some changes I propose. Please see the inline comments.

goto40:
> I like the addition of `get_reference_propositions` as it could be helpful for further tooling development (/cc @danixeee).

The additions concerning the editor support will be part of a separate proposition. --> The possible new API will return names, not objects... (to be discussed separately!)

docs/scoping.md
goto40: I am sorry for the errors in the rendered version. After corrected the rendering I cleaned up the text. I think it is ok to have no separate section, since with the correctly rendered version you can see all relevant classes on one bullet point depth.

(text adapted)

CHANGELOG.md
goto40: text added.

tests/functional/examples/test_hierarchical_data_structures_referencing_attributes.py
goto40: Done.

textx/scoping/providers.py
goto40: This is a check for teh correct config of the scope provider.

It is expected to point to a list (of elements).
 * Replaced assert with if+raise
 * Added test
 * Added comment

textx/scoping/providers.py
goto40: Ok. Code removed.

textx/scoping/providers.py
goto40: Ok. code removed.

textx/scoping/providers.py
goto40: (same answer; no extra test added)

textx/scoping/providers.py
goto40: code removed.

goto40: and code changed (check for list); see last comment

textx/scoping/tools.py
goto40: code changed

textx/scoping/tools.py
goto40: Code changed.

textx/scoping/providers.py
goto40: Corrected scoping.md (removed "typically a list" --> "a list")

goto40:
I am done with the corrections defined by the review comments, so far.
I would be happy, if you could have a look at the changes..

igordejanovic:
Looks much better. 

There is still an issue with `get_referenced_object`. The name implies that a single object is returned (`object` in the name is singular). The docstrings at the beginning says:
```
get objects based on a path
```
which implies that multiple objects are returned, but in the `Returns:` section:
```
    Returns:
        the object if found, or Postponed() if some postponed
        refs are found on the path / or obj is not found
```
we again have a suggestion that a single object is returned.

Looking at the code it seems that list can be returned also. So this has to be cleared up as it is confusing.

goto40:
(At the end I can squash and merge and rewrite one single proper commit message)

goto40:
You are absolutely right. The function name is misleading and the docu as well. 
 * I adapted the documentation and clarified the responsibility.
 * I renamed the function to a much simple name to "resolve_model_path", because this is what happens.
 * I added some unittests to the existing ones.
   * I found a **bug** I "added" while refactoring: I moved a function argument to the wrong place - and it was covered by the existing tests. **Tests are now adapted** to check this case and the bug is **resolved**.
   * I did not cover the Postponed case (difficult to isolate), but I think most other cases are covered.

igordejanovic:
Thanks. It is *much* cleaner now!

> (At the end I can squash and merge and rewrite one single proper commit message)

Yes, please do. Or do interactive rebase and make it in several commits with nice log messages. However you see is better.

igordejanovic (approved):

Issue #193, opened 2019-05-29
Object processor execution order


goto40:
Somehow it seems we lost a change in the docs:
https://github.com/textX/textX/pull/184/files

In addition to this, we could allow to specify whether an object processor is executed before or after scope resolution, or automatically (default behavior, see missing docs in the referenced PR above).

This would allow users to handle cases like the ones mentioned in https://stackoverflow.com/questions/56362019/textx-how-to-generate-object-names-with-objectprocessors

I do not see an 'ultimate answer' to the question whether a processor has to be executed before or after scope resolution. In general, users will expect all references to be resolved (current state), but if the name of an object is modified, the other way round may be better...

A possible technical solution to 'tag' processors to be executed before or after could be a base class... Without base class, default behavior is selected.

igordejanovic:
> Somehow it seems we lost a change in the docs: https://github.com/textX/textX/pull/184/files

Very strange. Although the GitHub interface shows that the branch was merged it was left hanging unmerged. I've just merge it and publish the docs.

> I do not see an 'ultimate answer' to the question whether a processor has to be executed before or after scope resolution. In general, users will expect all references to be resolved (current state), but if the name of an object is modified, the other way round may be better...

I'm not sure that we should officially support such very specific use cases (where name gets changed in object processors). In those cases user should choose what (s)he feel is the right approach and implement it accordingly. As a last resort resolution can always be handled by a custom resolver that runs over the fully constructed model at the end.

goto40:
For me it sounds useful to be able to transform the model prior to scope resolution : we do the same for object processors working on match rules.

The feature has a clear use case (transform the model prior to scope resolution), a defined interface (e.g., based on a base class), and is not a BIC.

Name adaptations could be an additional valid use case, as well...

igordejanovic:
Ok. Sounds good. If you see a clear path to implement it in a backward compatible way that would not increase complexity and you would like to work on it then go ahead :)

goto40:
Alternatively, the problem could be solved with a user defined class: the name is deduced when constructing the element...

That would be even simpler...

goto40:
I will give it a try and the see what to propose to the user on SO...

igordejanovic:
> Alternatively, the problem could be solved with a user defined class: the name is deduced when constructing the element...

That would definitely be simpler although user would have to provide user classes for all model object where name change/definition is necessary.

goto40:
## Problem

[...]

PR #194, opened 2019-06-02
Name resolution logic for specific scope provider cases

goto40:
## Code review checklist

- [x] Commit messages are meaningful (see [this][commit messages] for details) - **squash**

Additional aspects:

- [x] Some tests involving importUri or Global repos.
- [x] some test checking fqn cases with the same name on different levels. Also with (artificially?) postponed reference resolutions. 

This is a pull request to **discuss** a possible solution for #193.

Playing around with this absolutely interesting issue revealed new aspects to me:

 *  **names dependent on model contents** (involving unresolved references). This **name resolution, which can be Postponed**, in terms of our reference resolution logic.
 * Even more interressing are the consequences of that: What happens to **references pointing to locations, where unresolved names are found**? Here, we must Postpone the reference resolution process, because we cannot know if the name might match when resolved...

In this pull request we propose a possible solution to #193. The solution is **very restricted**, in term of **where custom name resolution** can be applied (e.g. not for intermediate names in a FQN query, such as a package name in ```pack1.pack2.class1```; see docs).

With **TEP-001** https://github.com/textX/textX/wiki/Reference-resolving-expression-language-(RREL)
we might find much better ways to include custom name resolution, since the search for elements in the model tree is much more guided and thus restricted: It is exactly this unrestricted search, that makes a general custom nameing (including Postponed names) so difficult to handle for the end user (for a FQN based scope provider; for RelativeName/RelativeNameExt there is not such problem).
 
**(A)** We have to decide if we want to take this PR into account - or  - **(B)** to close it without merge and reference it from TEP-001.
 * **(+A)** Code changes are modarate (some tests are missing)
 * **(+A)** Code changes are very local and the effect is documented.
 * **(+B)** But: the functionality is possibly difficult to grasp for the end user, because of the restrictions we selected are not clear (without reading this PR and the docs).


docs/scoping.md
```diff
@@ -180,6 +180,44 @@ We provide some standard scope providers:
    to model inheritance or chained lookups**.
    Example: see [tests/test_scoping/test_local_scope.py](https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_local_scope.py).
 
+### Custom name resolution
+
+In some cases, the name of an element needs to be derived from some model data (instead of
+being specified explicitly via a ```name``` attribute). The FQN based scope providers and
```
igordejanovic: A single backtick is used for inline code. Triple backticks are used for code blocks. There are multiple instances in this docs section to change.

docs/scoping.md
```diff
@@ -180,6 +180,44 @@ We provide some standard scope providers:
...
+the name of an object** and also to **postpone the name resolution**, if the name depends 
+on unresolved references.
+
+ * The FQN scope providers make a type unspecific search of named containers to guide their
+   way through the model. To **prevent circular deperendencies** (of references waiting for
+   a postponed name to be resolved), we **restrict the name resolution logic to be 
+   applied to the final element only of a FQN reference string**: E.g., for 
+   ```"pack1.pack2.class1"```, the name resolution logic will only 
+   be applied to the final element (```"class1"```). 
+ * The RelativeName and the RelativeNameExt scope providers do not need such a restriction,
```
igordejanovic: `RelativeName` and `RelativeNameExt` are class names and should be surrounded by backticks.

docs/scoping.md
```diff
@@ -180,6 +180,44 @@ We provide some standard scope providers:
...
+   since they operate on a small set of reference names.
+
+**Functionality:** When reference resolution searches for a name for which the scope resoltion
```
igordejanovic: Typo: resoltion -> resolution

docs/scoping.md
```diff
@@ -180,6 +180,44 @@ We provide some standard scope providers:
...
+logic is applied, **all non-postponed names are taken into account**. 
+
+ * **(1)** If postponed names are found,
+   the reference to be resolved it Postponed.
+ * **(2)** Else, if a match is found, that object is used. 
+ * **(3)** Else, no match is found on that level.
+
+Custom name resolution is not available for PlainName based scope providers.
```
igordejanovic: `PlainName` is a class name.

textx/scoping/providers.py
```diff
@@ -182,24 +211,40 @@ def find_obj(parent, name):
...
                         for innerobj in obj:
-                            if hasattr(innerobj, "name") \
-                                    and innerobj.name == name:
-                                return innerobj
+                            if cls is None or\
+                                    textx_isinstance(innerobj, cls):
+                                if the_name_resolver_logic(innerobj)\
+                                        is not None:
+                                    myname =\
+                                        the_name_resolver_logic(innerobj)
```
igordejanovic: `the_name_resolve_logic` is called two times unnecessary.

textx/scoping/providers.py
```diff
@@ -15,10 +15,29 @@
 textx.metamodel meta model.
 
 See docs/scoping.md
-
 """
 
 
+def default_name_resolver_of_model_object(obj):
```
igordejanovic: Is there a reason why we wouldn't use a shorter name `default_name_resolver`?

textx/scoping/providers.py
```diff
@@ -410,7 +456,8 @@ def my_scope_redirection_logic_def(obj):
         else:
             my_scope_redirection_logic = scope_redirection_logic
         ImportURI.__init__(self, FQN(
-            scope_redirection_logic=my_scope_redirection_logic),
+            scope_redirection_logic=my_scope_redirection_logic,
+            name_resolver_logic=name_resolver_logic),
```
igordejanovic: Maybe now is harder to change in order to stay consistent with `scope_redirection_logic` but I think that the name `name_resolver` would be much nicer. `logic` suffix sounds superfluous. 

I would nevertheless go with `name_resolver` and plan to change `scope_redirection_logic` to `scope_redirection` or better yet `scope_redirector` in the future.

`name_resolver`and `scope_redirector` would be consistent with `importURI_converter`

textx/scoping/providers.py
```diff
@@ -524,6 +574,28 @@ def __init__(self, filename_pattern=None, glob_args=None):
...
+        lambda x: (name_resolver_logic(x), x),
+        obj_list))
+
+    name_obj_tuples = list(filter(
+        lambda x: textx_isinstance(x[1], cls) and (
+                type(x[0]) is Postponed or
+                x[0].find(name_part) >= 0),
+        name_obj_tuples))
+
+    return name_obj_tuples
```
igordejanovic, line 587: If I understand previous code correctly it could be made faster by filtering by `textx_isinstance` in the first assignment. Also, no need to convert to `list` in the first assignment. Is converting to list necessary in the second? I.e. do code that uses this function requires a list or can it just use iterable.

textx/scoping/providers.py
```diff
@@ -524,6 +574,28 @@ def __init__(self, filename_pattern=None, glob_args=None):
...
+
+
+def _filter_postponed_from_tuples_of_name_and_objects(
+        name_obj_tuples):
+    return list(filter(lambda x: type(x[0]) is not Postponed,
+                       name_obj_tuples))
```
igordejanovic, line 593: Same here. Is converting to list necessary? Lazy iterables are preferred for performance reasons.

textx/scoping/providers.py
```diff
@@ -572,18 +645,23 @@ def get_reference_propositions(self, obj, attr, name_part):
...
-        if type(lst) is Postponed:
-            return lst
-        if len(lst) > 0:
-            return lst[0]
+        lst_raw = self.get_reference_propositions(
+            obj, attr, obj_ref.obj_name)
+        if type(lst_raw) is Postponed:
+            return Postponed()
+        lst = _filter_postponed_from_tuples_of_name_and_objects(lst_raw)
+        if len(lst_raw) > len(lst):
```
igordejanovic, line 659: I see here that the list is necessary

igordejanovic (commented):
Overall looks good. I have some suggestions.

I don't mind this merged if you are for it. But I don't mind to leave this as a separate branch and focus on better solution during work on TEP-1 either. There are pros and cons for any of these two decisions.

You decide :)

docs/scoping.md
goto40: corrected.

docs/scoping.md
goto40: corrected

docs/scoping.md
goto40: corrected

docs/scoping.md
goto40: corrected

textx/scoping/providers.py
goto40: very good. An implementation will intuitively expect to be called once per element...

goto40: corrected

textx/scoping/providers.py
goto40: Ok. Corrected. Also adapted docs.

textx/scoping/providers.py
goto40: * Changed to `name_resolver`
 * Adapted docs
 * created ticket for the scope redirector (#195)

textx/scoping/providers.py
goto40: Note: Once we drop python 2.7, we may be able to change some of of the "get_list_of" to (recursive) generators. I think there are restrictions in 2.7 for that (recursive generators)...

goto40: This code could be optimized. It is implemented that way to be consistent (+somehow readable) with the spec (in scoping.md).

textx/scoping/providers.py
goto40: (No action, see comment below)

textx/scoping/providers.py
goto40: (No action, see comment below)

goto40:
Thank you for the detailed review. 

I wait some time if AGK (from SO) thinks the change is useful and clear for him.

goto40:
If we have no use case for this I wait for TEP-001...

goto40:
To clean up: We could close this PR. The branch ist still existing...

igordejanovic:
Ok. I agree. The PR will be there as closed as a reference.

PR #197, opened 2019-06-16
added new test function to check if a file is included by a model.

goto40:
## Code review checklist

**Purpose**: make it simple to check whether a filename is inlcluded by a model...

**Solution**: added a wrapper function.

**Details**: see scoping.md + test_model_repository.py

**Note**: I added some (more) "abspath" in some locations - I do not see that we get a BIC with this...

### DIFF
--- a/docs/scoping.md
+++ b/docs/scoping.md
@@ -197,6 +197,20 @@ have unique model elements (no double instances).
 Examples see [tests/test_scoping/test_import_module.py](https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_import_module.py).
 
 
+### Included model retrieval
+
+When a model includes or references other model (model files), the scope 
+providers in `textx.scoping.providers` use the field `_tx_model_repository`
+of the model object to keep track of the included models.
+
+ * You can get a list of all included model objects of a model
+   and the model itself with
+   `textx.scoping.providers.get_all_models_including_attached_models`.
+ * You can check if a model file is included by a model with
+   `textx.scoping.providers.is_file_included_by_model`.
+ 
+
+
 ## Technical aspects and implementation details
 
 The scope providers are Python callables accepting `obj, attr, obj_ref`:
--- a/textx/scoping/__init__.py
+++ b/textx/scoping/__init__.py
@@ -254,6 +254,22 @@ def get_all_models_including_attached_models(model):
     get a list of all models stored within a model
     (including the owning model).
 
+    @deprecated (BIC): use model_object.get_included_models()
+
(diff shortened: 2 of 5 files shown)


docs/scoping.md
```diff
@@ -197,6 +197,20 @@ have unique model elements (no double instances).
...
 
+### Included model retrieval
+
+When a model includes or references other model (model files), the scope 
+providers in `textx.scoping.providers` use the field `_tx_model_repository`
+of the model object to keep track of the included models.
+
+ * You can get a list of all included model objects of a model
+   and the model itself with
+   `textx.scoping.providers.get_all_models_including_attached_models`.
```
igordejanovic, line 208: I would prefer a shorter name for this function especially cause it has `attached` in its name while we haven't defined what `attached` means. I think it would be just enough for the name to be `get_included_models` or if more precision is needed (although I still prefer shorter names :) ) the name could be `get_model_with_included_models`.

textx/scoping/__init__.py
```diff
@@ -266,3 +268,22 @@ def get_all_models_including_attached_models(model):
     else:
         models = [model]
     return models
+
+
+def is_file_included_by_model(filename, model):
```
igordejanovic: Maybe it would be better to have `is_file_included` on `model` object?

docs/scoping.md
igordejanovic: `get_included_models` could be a method on a model object. See comment bellow for `is_file_included`

igordejanovic (commented):
Looks good. Just a few suggestions about naming. Please see inline comments.

textx/scoping/__init__.py
goto40: I like the idea. Where do you think is a good place to add the method. We do not have a "class Model" or so... Where do you suggest to add those methods (or attributes)?

igordejanovic: I was thinking about adding them like we add special attributes on the constructed model but after second thought I don't think that would be a good idea. It would pollute the namespace of user model object and adding them with `_tx_` prefix would be plain ugly.

goto40: I just had the same idea. I will keep free functions instead.

docs/scoping.md
goto40: Done. As free function.

textx/scoping/__init__.py
goto40: Done

goto40:
@danixeee : I will merge this branch if you have no further comments.

Issue #196, opened 2019-06-14
model_from_str: Allow passing file_name argument


danixeee:
This is needed for VS Code extension.

When model file is opened in VS Code, server receive events on document changes (text) and that is a reason why I need to call `model_from_str`. Calling `model_from_file` would work if document is saved, but we want to validate model while changing it.

I need either to pass `file_name` (file path) as an argument to method `model_from_str` or to call `model_from_file` with pre-loaded file. I saw that implementations for these two methods are slightly different, so I didn't go into details.

If I set [model._tx_filename](https://github.com/textX/textX/blob/master/textx/scoping/providers.py#L328) manually (while debugging) it works fine.

`textx_ls_core` validation logic is [here](https://github.com/textX/textX-LS/blob/danixeee/textx-view-container/textX-LS/core/textx_ls_core/features/validate.py#L13)

goto40:
A first workaround could be to use `pre_ref_resolution_callback` from `model_from_file` to set the filename... But we should analyze your issue in detail, because it is a fundamental need of an editor on my eyes.

I am not sure if the workaround is safe for multi file models...

igordejanovic:
@danixeee It seems that we already have that. See [this](https://github.com/textX/textX/blob/master/textx/metamodel.py#L57). `model_from_file` is just a thin wrapper around `model_from_str`. Have you tried to pass in `file_name` param?

@goto40 
> I am not sure if the workaround is safe for multi file models...

You are right. Even without workaround we should check if this will work in multi file models (errors, positions ...)

danixeee:
> See this.

@igordejanovic You referenced `metamodel_from_str`, but yes, it would be enough to do the same for [model_from_str](https://github.com/textX/textX/blob/master/textx/metamodel.py#L548). [parser.get_model_from_str](https://github.com/textX/textX/blob/master/textx/metamodel.py#L560) already has `file_name` parameter.

@goto40 I will try the workaround, thanks!

igordejanovic:
@danixeee Ah, yes, sorry. I believe that `get_model_from_str` should do the trick.

danixeee:
`**kwargs` [here](https://github.com/textX/textX/blob/master/textx/metamodel.py#L641) would be helpful, also.

I am having some issues when parsing models which use providers.

Model is opened and validated for the first time:

![image](https://user-images.githubusercontent.com/16227576/59539030-0a773c80-8efc-11e9-9ed9-8176451bca19.png)

[...]

PR #198, opened 2019-06-18
fix of issue 196: exception safety (obj processors do not report validation errors when reloading some models):

goto40:
* Problem was, that **with a global repository in the metamodel**, that **models with errors were cached** and reused (w/o beeing checked). Related docs: http://textx.github.io/textX/stable/scoping/#note-on-uniqueness-of-model-elements-global-repository
 * Solution: **delete all models touched while loading a model, when an error occurs** in object processors (in all repositories). - - > **strong exception safety guarantee**. 
 * _Alternative (not implemented)_: re-check cached models (Problem: what happens if the object processor modifies the model? Not good running object processors twice)
 * Minor: added (commented) printfs and commented some old printfs found in the codebase.

In principle this is a BIC. However, the previous behavior is a bug in my eyes... @igordejanovic: How to proceed?

## Code review checklist

- [x] Tests have been included and/or updated (**used testcase sketched in #196**, thank you @danixeee )
- [x] Docstrings have been included and/or updated, as appropriate **N/A**
- [x] Standalone docs have been updated accordingly **N/A??**

### DIFF
--- a/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_IMPORT_ERROR.eflow
+++ b/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_IMPORT_ERROR.eflow
@@ -1,0 +1,6 @@
+#include "data_structures_including_error.edata"   // <--- types must be lowercase
+#include "types.etype"
+
+algo A1 : Point -> City
+algo A2 : City -> Population
+connect A1 -> A2
--- a/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_IMPORT_SYNTAX_ERROR.eflow
+++ b/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_IMPORT_SYNTAX_ERROR.eflow
@@ -1,0 +1,6 @@
+#include "data_structures.edata"
+#include "types_with_syntax_error.etype"
+
+algo A1 : Point -> City
+algo A2 : City -> Population
+connect A1 -> A2
--- a/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_SYNTAX_ERROR.eflow
+++ b/tests/functional/registration/projects/flow_dsl/tests/models/MODEL_WITH_SYNTAX_ERROR.eflow
@@ -1,0 +1,6 @@
+UNEXPECTED_#include "data_structures_including_error.edata"
+#include "types.etype"
+
+algo A1 : Point -> City
+algo A2 : City -> Population
+connect A1 -> A                                     // <--- Unknown object "A" of class "Algo"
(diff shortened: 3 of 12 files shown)


danixeee:
There is still bug here. Please try this test:

```python
MODEL_WITH_IMPORT_ERROR = """
#include "data_structures_including_error.edata2"   // <--- types must be lowercase

#include "types.etype2"

algo A1 : Point -> City
algo A2 : City -> Population
connect A1 -> A2
"""

MODEL_WITH_TYPE_ERROR = """
#include "data_structures_including_error.edata2"
#include "types.etype2"

algo A1 : Point -> City
algo A2 : City -> Population
connect A1 -> A                                     // <--- Unknown object "A" of class "Algo"
"""


def test_flow():
    mmF = metamodel_for_language('flow-dsl-s')

    errors = []

    try:
        mmF.model_from_str(MODEL_WITH_IMPORT_ERROR)
    except TextXError as e:
        errors.append(e) # Caught: "types must be lowercase" (CORRECT)

    try:
        mmF.model_from_str(MODEL_WITH_TYPE_ERROR)
    except TextXError as e:
        errors.append(e) # Caught: "Unknown object "A" of class "Algo"" (CORRECT)

    try:
        mmF.model_from_str(MODEL_WITH_IMPORT_ERROR)
    except TextXError as e:
        errors.append(e) # Caught: "Unknown object "A" of class "Algo"" (WRONG)

    assert errors[0] == errors[2]
```

goto40:
Thank you for the extended example. I found the error. I need more time to analyse other (similar) cases. I will check model cleanup on errors...

Your case seems to work now (**reason/explanation of your error/observation**: It was not a obj_processor error, but a scope provider error....).

goto40:
Fixed behavior: We now clean up when
 * errors occur during parsing of included models,
 * during reference resolution, and
 * during object processor execution (as before)

Added tests.

goto40:
@danixeee: Sorry for all these inconveniences. But I did not have the objective to continue to work with models or meta-models after an exception occurred... I fear you will get some other effects or bugs with all that logic (especially with cached models, when updating some files...).

goto40:
Note: we now should have the **strong exception safety guarantee** for the activity 'load model'. See https://en.m.wikipedia.org/wiki/Exception_safety

danixeee:
@goto40 Just tried it now again and it works fine, thanks again for quick fix!

> Sorry for all these inconveniences. But I did not have the objective to continue to work with models or meta-models after an exception occurred...

One option was to reload meta-model every time before parsing, and that would happen on every key-press, so I think reusing meta-model is much better.

goto40:
@danixeee 
> One option was to reload meta-model every time before parsing, and that would happen on every key-press, but I think reusing meta-model is much better.

I see a whole bunch effects which could happen when updating a model. Compare #199. 

For a first implementation it would be nice to reload the meta model every time you enter/leave the current editor widget (on_focus_changed or similar?). This forces to reload everything...

goto40:
Close in favor of #200

Issue #199, opened 2019-06-21
Enhancing model caching to automatically reload model when they changed (e.g., on disc)


goto40:
I would like to enhance model caching to allow to automatically reload models (e.g., from disc) when models changed (e.g., on disc).

Need further analysis to understand how to allow a seamless integration into the (scope provider) logic of different predicates. Possible predicates:
 - date of file changed
 - MD5 of file changed
 - in-memory representation within an editor changed (?)

Also allow to update models.

**Example use case / unittest** with global repository:
 - load A,B,C (A is including B and C)
 - reload A (B,C are cached)
 - change B on disc
 - reload A (C is cached, B is reloaded)

**Example use case / unittest** with global repository:
 - load A,B,C (A is including B and C) + validate A,B,C
 - reload A (B,C are cached) + validate A (B,C are not validated because they are cached)
 - update B with update mechanism + validate B; --> also reload (request reload?) of all model including B

igordejanovic:
Good idea but not sure where this should fit. Should it be done in the core textX or should we do that in textX-LS? We should definitely have some generic extension mechanism in the model loading/caching process that could be used to implement this. There are many ways we could detect model changes. By pooling date/MD5 as you suggested or by using some platform specific feature like [inotify](https://en.wikipedia.org/wiki/Inotify) for example.

goto40:
I had the idea to provide some callback mechanism to hook-in the scope provider logic caching files. This hook could be implemented by textx-ls.

Let's think about it...

goto40:
If you have ideas, we can collect requirements / use cases here... (for the new mechanism).

danixeee:
I have already started implementing this in `textX-LS` (not yet completed).
When textX language project is installed in _editable_ mode, file system watcher is registered for its project path. When `.tx` files are changed extension is regenerated and updated (see [this](https://github.com/textX/textX-LS/blob/danixeee/textx-view-container/client/src/services/projectService.ts#L160)). I need to add watcher for `.py` files as well, and reload meta-models for already opened documents (no need to update extension, because coloring depends just on grammar keywords).

Thoughts on that approach?

goto40:
Hm. Not sure.

If we can handle it completely within textX-ls, perfect. Maybe we need support from the textx core api.

Check the use cases above. We should define what should happen when files and included files are modified. When to reload files...

I think we may need an api call to invalidate cached files... Maybe the new `get_included_files` api function is enough?

goto40:
[...]

PR #200, opened 2019-06-21
issue196 exception safety, rebase

goto40:
* Problem was, that **with a global repository in the metamodel**, that **models with errors were cached** and reused (w/o beeing checked). Related docs: http://textx.github.io/textX/stable/scoping/#note-on-uniqueness-of-model-elements-global-repository
 * Solution: **delete all models touched while loading a model, when an error occurs** in object processors (in all repositories). - - > **strong exception safety guarantee**. 
 * _Alternative (not implemented)_: re-check cached models (Problem: what happens if the object processor modifies the model? Not good running object processors twice)
 * Minor: added (commented) printfs and commented some old printfs found in the codebase.

Rebased version - follow-up PR (see #198)

@igordejanovic to easily review the changes, enable "Hide whitespace changes" in your diff-viewer (e.g. github browser window). There is mainly one change (at two locations) in model.py to handle exceptions.

## Code review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,15 +16,15 @@ please take a look at related PRs and issues and see if the change affects you.
 
 ### Added
 
-  - Added new function `textx.scoping.is_file_included` [#197]
+  - Added new function `textx.scoping.is_file_included` ([#197]).
 
 ### Changed
 
   - Changed function name `textx.scoping.get_all_models_including_attached_models`
-    to `textx.scoping.get_included_models` [#197] (marked old function
-    as deprecated).
+    to `textx.scoping.get_included_models` (marked old function
+    as deprecated) ([#197]).
   - Delete all models touched while loading a model, when an error occurs 
-    while loading in all repositories (strong exception safety guarantee). [#200]
+    while loading in all repositories (strong exception safety guarantee) ([#200]).
 
 ## [v2.0.1] (released: 2019-05-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -409,6 +409,8 @@ please take a look at related PRs and issues and see if the change affects you.
   - Export to dot.
 
 
+[#200]: https://github.com/textX/textX/issues/200
+[#197]: https://github.com/textX/textX/issues/197
 [#188]: https://github.com/textX/textX/issues/188
 [#187]: https://github.com/textX/textX/pull/187
 [#186]: https://github.com/textX/textX/pull/186
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ please take a look at related PRs and issues and see if the change affects you.
   - Changed function name `textx.scoping.get_all_models_including_attached_models`
     to `textx.scoping.get_included_models` [#197] (marked old function
     as deprecated).
+  - Delete all models touched while loading a model, when an error occurs 
+    while loading in all repositories (strong exception safety guarantee). [#200]
 
 ## [v2.0.1] (released: 2019-05-20)
 
--- a/tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
+++ b/tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
@@ -61,7 +61,6 @@ def test_issue196_errors_in_scope_provider_and_obj_processor():
         _tx_model_repository.all_models.filename_to_model.keys())
 
 
-
 def test_issue196_syntax_error_1():
     mmF = metamodel_for_language('flow-dsl')
 
@@ -75,9 +74,10 @@ def test_issue196_syntax_error_1():
 
     with pytest.raises(TextXError,
                        match=r'.*error: Expected.*'):
-        mmF.model_from_file(os.path.join(current_dir,
-                                         'models',
-                                         'MODEL_WITH_IMPORT_SYNTAX_ERROR.eflow'))
+        mmF.model_from_file(
+            os.path.join(current_dir,
+                         'models',
+                         'MODEL_WITH_IMPORT_SYNTAX_ERROR.eflow'))
 
     # print("-----------------------------------2---")
     # print(metamodel_for_language('flow-dsl').
--- a/tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
+++ b/tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
@@ -10,12 +10,13 @@ current_dir = os.path.dirname(__file__)
 def test_issue196_example():
     mmF = metamodel_for_language('flow-dsl')
 
-    print("-----------------------------------1---")
-    print(metamodel_for_language('flow-dsl')._tx_model_repository.all_models.filename_to_model.keys())
+    # print("-----------------------------------1---")
+    # print(metamodel_for_language('flow-dsl').
+    #      _tx_model_repository.all_models.filename_to_model.keys())
 
     cached_model_count = len(
         metamodel_for_language('flow-dsl').
-            _tx_model_repository.all_models.filename_to_model.keys())
+        _tx_model_repository.all_models.filename_to_model.keys())
 
     with pytest.raises(TextXError,
                        match=r'.*types must be lowercase.*'):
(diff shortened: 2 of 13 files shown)


goto40:
@danixeee for your information, i switched the branch (rebase).

@igordejanovic  You could have a look when you find time (see text in PR on how to easily see the main change).

igordejanovic:
@goto40 Sorry, I just now noticed this PR. Somehow I didn't get a notification email from GH. Will do the review during the day.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ please take a look at related PRs and issues and see if the change affects you.
   - Changed function name `textx.scoping.get_all_models_including_attached_models`
     to `textx.scoping.get_included_models` [#197] (marked old function
     as deprecated).
+  - Delete all models touched while loading a model, when an error occurs 
+    while loading in all repositories (strong exception safety guarantee). [#200]
```
igordejanovic: #200 is missing in the list of PRs

tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
```diff
@@ -0,0 +1,116 @@
...
+                                         'MODEL_WITH_IMPORT_ERROR.eflow'))
+
+    # print("-----------------------------------2---")
+    # print(metamodel_for_language('flow-dsl').
+    #      _tx_model_repository.all_models.filename_to_model.keys())
+
+    # error while reading, no file cached!
+    assert cached_model_count == len(
+        metamodel_for_language('flow-dsl').
+        _tx_model_repository.all_models.filename_to_model.keys())
```
igordejanovic: `keys()` is superfluous. BTW, the comment says `no file cached`. Should we check that `len` returns `0`?
There are more instances of this bellow.

textx/model.py
```diff
@@ -598,92 +598,140 @@ def call_obj_processors(metamodel, model_obj,
...
+                # (remove all of them, not only the model with errors,
+                # since, models with errors may be included in other models)
+                for m in models:
+                    for m2 in models:
+                        if hasattr(m2._tx_metamodel, "_tx_model_repository"):
+                            m2._tx_metamodel.\
+                                _tx_model_repository.remove_model(m)
+                        if hasattr(m2, "_tx_model_repository"):
+                            m2._tx_model_repository.remove_model(m)
+                raise e
```
igordejanovic, line 697: This code could be simplified if we introduce `remove_models` which accepts an iterable of models. Then this can be reduced to a simple loop (no nesting). Another issue is that this code is maintaining the cache which is not a concern of `model.py` module but scoping/multi-model support. I suggest making an API in scoping for dealing with this issue and just calling that api from here. That API should also provide a function that would invalidate the cache given  an iterable of changed model paths we discussed in #199.

The same comment holds for the next except block.

igordejanovic (commented):
@goto40 Looks good. I have some suggestions. Please see inline comments.

CHANGELOG.md
goto40: Done.
Also added #197

tests/functional/registration/projects/flow_dsl/tests/test_issue196.py
goto40: Removed `keys()`.
Added comment to clarify that the count is checked to be unchanged (and not 0).

textx/model.py
goto40: 1. I created **helper functions** to remove models from the cache (and **removed all nested loops**).
2. I simplified the code (I omitted one superfluous deletion).

However, **I did not** introduce the **filename based invalidation** of models. For this I prefer having a look at all aspects of Repositorys in #199 at once and make refactorings (e.g.,  making the map inside the repositories private). 

Here, I **remove models based on the model, not on the file name**. I prefer this here (in principle, I can have "anonymous" entries, see https://github.com/textX/textX/blob/master/textx/scoping/__init__.py#L210, even this should not be possible when including models in our case).

@igordejanovic Maybe you can have a second look here...

goto40: Note: I just fixed a bug (only reported for python 3.4 - our luck that python 3.4 has another way to iterate through dicts). --> I reused a filter-object (empty after the first usage) --> solved by converting to a list.

textx/scoping/__init__.py
```diff
@@ -305,3 +330,22 @@ def is_file_included(filename, model):
         return all_entries.has_model(filename)
     else:
         return False
+
+
+def remove_models_from_repositories(model, models_to_be_removed):
+    """
+    Removed models from all relevant repositories (_tx_model_repository
```
igordejanovic: Removed -> Remove

textx/scoping/__init__.py
```diff
@@ -305,3 +330,22 @@ def is_file_included(filename, model):
...
+    of models and metamodel, if applicable).
```
igordejanovic: of models -> of model? We are removing from `_tx_model_repository` of the `model` param, right?

textx/model.py
```diff
@@ -598,92 +599,129 @@ def call_obj_processors(metamodel, model_obj,
...
+                                                      reverse=True))
+    # exception occurred during model creation
+    except BaseException as e:
+        # remove all models beeing constructed
+        all_affected_models = get_included_models(model)
+        models_to_be_removed = list(filter(
+                lambda x: hasattr(x, "_tx_reference_resolver"),
+                all_affected_models))
+        for m in all_affected_models:
+            remove_models_from_repositories(m, models_to_be_removed)
```
igordejanovic: This block of code has only `model` as it input and the comment says what it is doing. It seems to me that the block could be made into a function in scoping/multi-model support and just called here like (notice no comment is necessary):
```python
except BaseException as e:
    remove_all_affected_models_in_construction(model)
    raise e
```
Is there any reason to keep this piece of code here? 
I have a feeling that this piece of functionality could be reused elsewhere. Maybe I'm wrong.
The only reason I could think of is the line `model._tx_reference_resolver = None` which marks the model being constructed above which is in connection to this block of code and thus removing this block of code would somehow call for that line above to be "removed" or hidden but I don't see an obvious solution. Maybe:
```python
start_model_construction(model)
```

textx/model.py
igordejanovic: @goto40 Looks much cleaner. However, maybe it could be replaced by a single API call. I'm not totally sure that would be a good approach here. Please see the comment for the next except block bellow. It just feels like this might be a functionality that could be reused in other places in the future and thus could be beneficial to be made a function call in scoping/multi-model.

igordejanovic: @goto40 I agree for file name based invalidation to be part of #199. It should be carefully analyzed and implemented to be easy for usage from textX-LS.

igordejanovic (commented):
@goto40 Thanks. Looks much better. I have some more suggestions. Please see inline comments.

textx/scoping/__init__.py
goto40: done.

textx/scoping/__init__.py
goto40: Yes. Done. Also clarified the first argument description.

textx/model.py
goto40: Very good point. In order to increase the readability of the code I introduced `_start_model_construction` and `_remove_all_affected_models_in_construction` as suggested.

Because the decision of using the reference_resolver as "flag" for model to mark them as "being constructed" I left the two function in model.py as private functions. These can later easily be refactored...

I you prefer them somewhere else, I can move them.

textx/model.py
goto40: Single API call realized by allowing to remove from a list of models in `remove_models_from_repositories` (instead of from a single model).

textx/scoping/__init__.py
goto40: (obsolete, see other comment above/below regarding the "single function call")

textx/model.py
goto40: I also added `_end_model_construction`...

goto40:
Thank you for the comments. I reworked all commented code changes.

igordejanovic:
@goto40 Good work! No more comments. Please cleanup branch history and merge (or just squash merge).

PR #209, opened 2019-09-20
issue205 skipws propagation (regression tests)

goto40:
Here, we add some tests to check and demonstrate the propagation of skipws/noswkips across rules.

We should consider to add this tests, once the Arpeggio fix for the Arpeggio issue 61 is done.

## Code review checklist

- [n/a] Commit messages are meaningful (see [this][commit messages] for details) - squash
- [n/a] Changelog(s) has/have been updated, as needed (see `CHANGELOG.md`).

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -824,6 +824,10 @@ specified in brackets (`[ ]`) at the beginning of the rule's definition after
 the rule's name. Currently, they are used to alter parser configuration for
 whitespace handling on the rule level.
 
+Rule modifiers act on the current rule and all rules referenced inside the rule 
+(recursively): unless a refrenced rule has an explicit rule modifier, the currently
+active modifier state is propagated to referenced rules.
+
 There are two rule modifiers at the moment:
 
 * **skipws, noskipws** - are used to enable/disable whitespace skipping during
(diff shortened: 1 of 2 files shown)


igordejanovic:
@goto40 Arpeggio bugfix release 1.9.1 is out with a regression test for the issue. 

I don't think that we also need regression tests in textX as well as it was solely an Arpeggio issue but I suggest, since you have spent time to investigate `skipws` behavior, to move some of these tests (I would prefer a non-redundant subset) into standard battery of tests to cover `skipws` better.

goto40:
OK. Great. I will rework the test cases.

goto40:
* I reworked the tests and the documentation alongside the test cases.
 * I added some explanation to the docs (grammar.md)

Feel free to comment...

**However**: I installed the previous version of Arpeggio (```git co aff1bb8```) and all tests pass. I cannot (re)produce an error showing the new Arpeggio version has an effect... Any idea, what to do to demonstrate the old Arpeggio version was a problem for textx?

igordejanovic:
@goto40 Created a regression test that fails with the unfixed version of Arpeggio and pass with fixed version. I think that this single test is all what we need. I've reused your docstrings for explanation in comment. If you agree with the change I would merge/cherry pick just this single commit and your addition to the docs.

goto40:
Very good! Good work.
OK for me, go ahead as you described.

Issue #176, opened 2019-03-30
Release process


igordejanovic:
It is a time to define some lightweight process that will govern textX development and the release process.

Here is a suggestion:
- We are moving to semantic versioning and a standard format to keep changelogs (see [CHANGELOG.md](https://github.com/textX/textX/blob/master/CHANGELOG.md)).
- We develop features on feature branches. Feature branches are named `feature/<feature name>`.
- We fix bugs on bugfix branches named as `bugfix/<issue no>-<name>`
- We have a branch for the upcoming major release -- `next-release` (red in the picture bellow).
- If the feature is backward incompatible (BIC) the PR is made against the `next-release` branch (not against the `master` branch)
- If the feature is backward compatible PR is made against the `master`.
- Thus, `Unreleased` section in the changelog on the `master` branch will never have any BIC change. All BIC changes should go to the changelog on the `next-release` branch.
- We constantly merge `master` branch  to `next-release` branch. Thus, `next-release` branch is the latest and greatest version with **all** finished features applied.
- When the time for major release come we create maintenance branch from the last released version/tag of the form `<major>.<minor>.x` (green in the picture bellow), we merge `next-release` branch to `master` and  tag the new major version release.
  
Here is a drawing that depicts the process:
![image](https://user-images.githubusercontent.com/1673425/58026599-19deb200-7b17-11e9-9869-033f00889a5c.png)

This approach makes releasing minor and bugfix releases possible while developing BICs for the next major release. Users should be safe declaring a dependency towards the major version. Also, following `master` branch should be safe up until a new major release has been made.

goto40:
Very good!

How do we create feature/bugfix branches? Naming conventions?

danixeee:
When we discover and fix a bug that exists in _all_ major releases, how we are going to handle it?
I guess fixed bug can be merged to master and the lts release can be tagged with increased version (e.g. v2.4.2). For non-lts releases we will create a release branch from a tag (e.g. v1.2.3), update changelog, etc and release bugfix from there, am I right?

igordejanovic:
> How do we create feature/bugfix branches? Naming conventions?

We can use `feature/feature_name` and `bugfix/issue...`.

> When we discover and fix a bug that exists in all major releases, how we are going to handle it? I guess fixed bug can be merged to master and the lts release can be tagged with increased version (e.g. v2.4.2). For non-lts releases we will create a release branch from a tag (e.g. v1.2.3), update changelog, etc and release bugfix from there, am I right?

[...]

PR #214, opened 2019-10-12
Release process docs

igordejanovic:
Related to  #176 

These are release process docs. I'll check the release step using the current 2.1.0 release.

## Code review checklist


docs/process.md
```diff
@@ -0,0 +1,75 @@
...
+  12. Merge PR and delete PR branch (`release-preparation`).
+  13. Change the version in `textX/__init__.py` to next minor version with
+      `.dev0` addition (e.g. `v2.2.0.dev0`).
+  14. Merge `master` to `next-version` to keep it up-to-date.
+
+
+!!! note
+    For supporting previous versions only bugfix releases will be made. The
+    process is similar. The difference is that we base our release process of
+    the `support` branch instead of `master`. Thus we skip step 1, in step 5 we
```
goto40: I am not sure. Maybe it is clearer that way (+"here", changed:"on the master..."):

The difference here is that we base our release process on the `master` branch' instead  of the `support` branch.

goto40 (commented):
Sounds good so far. I like to make the release process clear and transparent (to allow anybody to understand what is released!) Very good work!

I am looking forward to try the editor support... @danixeee : I unfortunately did not have the time to follow all your activities the last weeks. But it sounds to me to be an impressive work! 

**Editor support for models** (given installed language meta models) will be a cornerstone of textx from an enduser point of view and will surely significantly increase the imapct of textx on other software tools. Great work!!

docs/process.md
igordejanovic: @goto40 Yes, I think this needs rewording. Done in the last commit. See if now sounds better.

docs/process.md
```diff
@@ -0,0 +1,86 @@
...
+      # mike delete stable  # this deletes alias stable
+      # mike deploy 1.9 stable   # and adds it to 1.9 minor version
+      mike deploy 1.9 stable -p
+      ```
+
+!!! note
+    For supporting previous versions only bugfix releases will be made. The
+    process is similar. The difference for support release would be that release
+    process would be based of the `support` branch instead of the `master`
+    branch as is done for regular releases. Thus we would skip step 1, in step 5 we
```
goto40: Did I udnerstand it correctly:
"Thus we would skip step 1, ..."?
-->
"Thus, **for regular releases,** we would skip step 1, ..."?

igordejanovic: No, that is for support releases. :) Updated wording to make that explicit.

goto40: sorry... OK for me.

goto40 (approved):
OK for me.

docs/process.md
igordejanovic: np, thanks for the review. It did sound ambiguous before

PR #224, opened 2020-01-26
Enable passing of metamodel params in mm registration API calls

igordejanovic:
In order to make calls `metamodel_for_file/language` symmetric to `metamodel_from_str/file` this adds keywords params to registration API calls that are used to configure the meta-models.

## Code review checklist

### DIFF
--- a/docs/registration.md
+++ b/docs/registration.md
@@ -383,20 +383,31 @@ All classes and functions documented here are directly importable from `textx` m
   parameters
 - `clear_language_registrations()` - deletes all languages registered
   programmatically. Note: languages registered through `setup.py` won't be removed
-- `metamodel_for_language(language_name)` - returns a meta-model for the given language name
+- `metamodel_for_language(language_name, **kwargs)` - returns a meta-model for
+  the given language name. `kwargs` are additional keyword arguments passed to
+  meta-model factory callable, similarly to `metamodel_from_str/file`.
 - `language_for_file(file_name_or_pattern)` - returns an instance of `LanguageDesc`
   for the given file name or file name pattern. Raises `TextXRegistrationError`
   if no language or multiple languages can parse the given file
 - `languages_for_file(file_name_or_pattern)` - returns a list of `LanguageDesc`
   for the given file name or file name pattern
-- `metamodel_for_file(file_name_or_pattern)` - returns a language meta-model for
-  a language that can parse the given file name or file name pattern. Raises
+- `metamodel_for_file(file_name_or_pattern, **kwargs)` - returns a language
+  meta-model for a language that can parse the given file name or file name
+  pattern. `kwargs` are additional keyword arguments passed to meta-model
+  factory callable, similarly to `metamodel_from_str/file`. Raises
   `TextXRegistrationError` if no language or multiple languages can parse the
   given file
 - `metamodels_for_file(file_name_or_pattern)` - returns a list of meta-models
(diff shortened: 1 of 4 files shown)


docs/registration.md
```diff
@@ -383,21 +383,31 @@ All classes and functions documented here are directly importable from `textx` m
   parameters
 - `clear_language_registrations()` - deletes all languages registered
   programmatically. Note: languages registered through `setup.py` won't be removed
-- `metamodel_for_language(language_name)` - returns a meta-model for the given language name
+- `metamodel_for_language(language_name, **kwargs)` - returns a meta-model for
+  the given language name. `kwargs` are additional keyword arguments passed to
+  meta-model factory callable, similarly to `metamodel_from_str/file`.
```
goto40, line 388: Good idea... but I am not sure about the consequences / use case: I just wondered if a language may have some restrictions, some decisions about meta model already made by the language designer. 

--> Is it a use case that I request a metamodel for a (predefined, ready-to-use) language with other meta model parameters that the language designer defined for me...?

docs/registration.md
```diff
@@ -383,21 +383,31 @@ All classes and functions documented here are directly importable from `textx` m
...
   for languages that can parse the given file name or file pattern
 - `language` - a decorator used for [language registration](#registering-a-new-language)
 
 
+!!! warning
+
+    meta-model instances are cached so it is cheap to call
+    `metamodel_for_file/language` multiple times but `kwargs` used in the
+    first call only will be applied as the further calls return cached instance.
+
```
goto40, line 411: Is it a good idea to *prevent using the cache* when passing some kwargs? 

This would allow to build the customized meta model once (with kwargs), then you can retrieve the registered language afterwards (w/o kwargs).

goto40 (commented):
Nice to hear from you. I have some comments. 

My overall feeling is good. I think the addition is ok. The most critical point is the caching effect, which might lead to hard-to-find bugs in my eyes...

docs/registration.md
igordejanovic: Good point. I found this use case when trying to built meta-model with a particular set of user classes and figured out that a need to override other stuff may be needed. Let's say that those kwargs should be whatever language designer is defined as overridable. Those params are passed to a meta-model callable factory so the factory can decide what to do with those.

Probably some more clarification needed in the docs.

docs/registration.md
igordejanovic: Great idea. That would prevent buggy behavior if user would make a call with different kwargs while still retaining good performance in general case. I'll rework it to reinit meta-model whenever kwargs is not empty.

docs/registration.md
goto40: This is ok. Since a meta model provider (language designer) can prevent such actions and throw and error if kwargs are passed (and not expected). 

After all, the language factory is fully in the hand of the language designer (or am I wrong?).

For me it is ok.

docs/registration.md
goto40: Great!

docs/registration.md
igordejanovic: Yes, you are right. The language designer can decide what to do with passed parameters.

igordejanovic:
@goto40 I've reworked the `kwargs` cache circumventing as per your suggestion and updated the docs and added tests. This is ready for another round of review.

goto40:
Good job!

goto40 (approved):

PR #225, opened 2020-01-26
Better dot visualization

igordejanovic:
Improved dot visualization of meta-models:

- rendering of base-types removed
- optional/required attrs rendering changed
- all match rules are rendered as a single table

If this looks better I'll update images in the docs.

## Code review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ please take a look at related PRs and issues and see if the change affects you.
   - Fixed return value of textx generate and check commands: we return a failure
     on error now ([#222])
   - Fixed type checking for references to builtin elements ([#218])
+  
+### Changed
+
+  - Dot rendering of meta-model: remove rendering of base types, improve
+    rendering of required/optional, render match rules as a single table.
+    ([#225])
 
 
 ## [v2.1.0] (released: 2019-10-12)
@@ -428,6 +434,7 @@ please take a look at related PRs and issues and see if the change affects you.
   - Export to dot.
 
 
+[#225]: https://github.com/textX/textX/pull/225
 [#222]: https://github.com/textX/textX/pull/222
 [#219]: https://github.com/textX/textX/pull/219
 [#218]: https://github.com/textX/textX/pull/218
--- a/examples/render_all_grammars/render_all_grammars.py
+++ b/examples/render_all_grammars/render_all_grammars.py
@@ -3,6 +3,7 @@ from textx import metamodel_from_file
 from textx.export import metamodel_export, PlantUmlRenderer
 import fnmatch
 import os
+import sys
 from os.path import sep, join, dirname, exists
 
 
@@ -67,4 +68,7 @@ def main(path=None, debug=False, reportfilename=None):
 
 
 if __name__ == "__main__":
-    main()
+    if len(sys.argv) == 2:
+        main(sys.argv[1])
+    else:
+        main()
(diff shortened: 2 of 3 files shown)


goto40:
Very nice!

I checked `examples_StateMachine_state_machine.dot.png` after running `python render_all_grammars.py`. Much easier to read.

Do you also plan to update the plantuml version (e.g. `examples_StateMachine_state_machine.png`)?

igordejanovic:
> Do you also plan to update the plantuml version (e.g. `examples_StateMachine_state_machine.png`)?

After looking at `pu` outputs I think we should reorder type/name. E.g. instead of `ID name` we should render `name: ID` which will be better aligned with both dot and standard UML class diagramming notation. What do you think?

I don't know if PlantUML has some similar feature that would enable rendering of match rules as a table.

goto40:
I played with the *.pu file manually: I added 
```
title
  Matched rules:
  |= Name  |= Rule details |
  | Comment | \\/\\*(.\|\\n)*?\\*\\/ |
  | Keyword | end\|events\|resetEvents\|state\|actions |
  | SMID | ID |
end title
```
to add a table as title (maybe we could also add it as a note). I added a backslash before all backslashes and pipe symbols.

The result is attached.
examples_StateMachine_state_machine

goto40:
> render `name: ID` which will be better aligned with both dot and standard UML class diagramming notation. What do you think?

Yes.

igordejanovic:
@goto40 Looks really good. Yes, maybe a note would be better. `Matched rules:` should be `Match rules`. Could you add your changes to this branch? Also, when your are on it please reorder the name/type.

goto40:
I am happy to do it

goto40:
I added a first version.
- I adapted the plantuml renderer in the same way as the dot renderer
- I reused the `dot_match_str` function for plantuml support
- I flipped name/type

To be reviewed carefully:
- I adapted the `dot_match_str` to handle, e.g. "Not" in `state_machine.tx`. **Maybe I am too generic here**... (else w/o type check; see diff to previous version).
- I had to pass the "known" match rules, in order to not expand them in the `dot_match_str` function. See image below, compared to the image above (usage of `Keyword` in the `SMID` rule). This fix was also applied for the dot export.

Any idea how we could design some good unittests here? Of course we could just make some regexp checks... Would be better than nothing. 

I have the feeling that the export function is not tested enough.

examples_StateMachine_state_machine
examples_StateMachine_state_machine dot

The grammar:
```
...
Keyword:
    'end' | 'events' | 'resetEvents' | 'state' | 'actions'
;
...
SMID:
    !Keyword ID
;
...
```

igordejanovic:
@goto40 Great work

> * I adapted the `dot_match_str` to handle, e.g. "Not" in `state_machine.tx`. **Maybe I am too generic here**... (else w/o type check; see diff to previous version).

Looks good. I think this approach is perfectly fine.

> Any idea how we could design some good unittests here? Of course we could just make some regexp checks... Would be better than nothing.
> 
> I have the feeling that the export function is not tested enough.

You are right, we are not covering this as we should. I don't have an idea what would be a good approach. I don't want to put something that will require frequent test updates. We can think about it and in the meantime we can do visual inspection over produced images.

I did some more testing and found two more problems. Fixed in the previous two commits. By running `render_all_grammars.py` over the whole code and visual inspection it looks very good to me.

goto40 (approved):
Very good!

Ok for me...

PR #231, opened 2020-02-12
Split CLI/dev tools to a separate project

igordejanovic:
This is bigger one and thus we'll merge against `next-release` as backward incompatible.

With this rework I'm trying to separate what is needed at development out of the main project. This repo will contain only the core features needed at run-time. The project that will host development features will be [textX-dev](https://github.com/textX/textX-dev). This project contains CLI commands, meta-model exporting functionality etc. 

The benefit is that in run-time, all the user will have to install will be textX and the only dependency will be Arpeggio. No `textx` cli command will be available. If the user install `textX-dev` (or install this repo with `dev` extras, e.g. by `pip install textX[dev]` than it will pull in `textX-dev` will all dependencies needed for development (e.g. click, scaffolding...).

This change will enable us to introduce new dev dependencies without worrying that we will have unnecessary overhead for deployment.

In addition I'm moving project's meta-data from `setup.py` to `setup.cfg`. Also, introducing `setuptools_scm` for managing versions using git history/tags.

## Code review checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,15 +82,13 @@ Ready to contribute? Here's how to set up `textX` for local development.
 
         $ git clone git@github.com:your_name_here/textX.git
 
-3. Install your local copy into a virtualenv. Assuming you have
-   [virtualenv](https://virtualenv.pypa.io/en/latest) installed, this is how you
-   set up your fork for local development:
+3. Install your local copy into a virtualenv. This is how you set up your fork
+   for local development:
 
         $ cd textX/
-        $ virtualenv venv
+        $ python -m venv venv
         $ source venv/bin/activate
-        $ pip install -r requirements_dev.txt
-        $ pip install -e .
+        $ ./install-dev.sh 
         
     Previous stuff is needed only the first time. To continue working on textX
     later you just do:
@@ -123,6 +121,10 @@ Ready to contribute? Here's how to set up `textX` for local development.
         $ py.test tests/functional/
         $ coverage run --source textx -m py.test tests/functional
         $ coverage report
+        
+   You can run all this at once with provided script `runtests.sh`
+   
+        $ ./runtests.sh
 
    In case you have doubts, have also a look at the html rendered version of
    the coverage results:
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ license = MIT
 description = Meta-language for DSL implementation inspired by Xtext
 keywords = parser, meta-language, meta-model, language, DSL
 url = https://github.com/textX/textX
-long_description = file: README
+long_description = file: README.md
 long_description_content_type = text/markdown
 classifiers =
     Development Status :: 5 - Production/Stable
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,61 @@
 [metadata]
+name = textX
+author = Igor R. Dejanovic
+author_email = igor.dejanovic@gmail.com
+license = MIT
+description = Meta-language for DSL implementation inspired by Xtext
+keywords = parser, meta-language, meta-model, language, DSL
+url = https://github.com/textX/textX
+long_description = file: README
+long_description_content_type = text/markdown
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    Intended Audience :: Science/Research
+    Topic :: Software Development :: Interpreters
+    Topic :: Software Development :: Compilers
+    Topic :: Software Development :: Libraries :: Python Modules
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
 description-file = README.md
+
+[options]
+packages = textx, textx.scoping
+zip_safe = False
+install_requires = Arpeggio>=1.9.0
+include_package_data = True
+package_dir =
+    textx = textx
+setup_requires =
+    setuptools_scm
+    wheel
+
+[options.extras_require]
+dev =
+    textx-dev
+    mkdocs
+    mike
+test =
+     textx-dev
+     flake8
+     tox
+     jinja2
+     coverage
+     coveralls
+     pytest
+
+[options.entry_points]
+textx_languages =
+    textx = textx.metamodel:textx
 
 [bdist_wheel]
 universal=1
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,26 +26,26 @@ classifiers =
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
-description-file = README.md
 
 [options]
-packages = textx, textx.scoping
+packages = textx, textx.scoping, textx.cli
 zip_safe = False
-install_requires = Arpeggio>=1.9.0
+install_requires =
+    Arpeggio>=1.9.0
+    click==7.0
 include_package_data = True
 package_dir =
     textx = textx
 setup_requires =
-    setuptools_scm
     wheel
 
 [options.extras_require]
 dev =
-    textx-dev
+    textX-dev
     mkdocs
     mike
+    twine
 test =
-     textx-dev
      flake8
      tox
      jinja2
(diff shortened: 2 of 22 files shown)


goto40:
If I understood correctly, the structure will be roughly as follows
  * textx / textx-core
    * metamodel.py/model.py/scoping/etc. stuff
    * registration (!)
  * textx-dev
    * export.py stuff
    * cli stuff
    * language scaffolding 

**I like the idea to separate the non-core functionality**, like the export stuff.

The only thing I am thinking of, is:
  * The registration functionality is fundamental to textx (because grammars depend on it),
  * but w/o a minimum command line interface I cannot query the installed languages.

Is this what we want? It is ok, but just a point we should make clear: when getting an error "language XY is not installed", we cannot query if XY is maybe misspelled... We also cannot check interactively the state of installed languages.

_For me, once I was familiar with the registration concept, the **textx-cli** became a **central tool**, making any other ugly quick-hacked custom "main-application" unnecessary!_ 

Of course, this could always be installed with textx-dev, only that a user of the language (using the textx cli) is not a language developer, but a language user! 

To sum up: 
 * I am slightly favoring the idea of leaving the core cli within the core textx. The click dependency is not a bad thing...
 * But I also understand the wish to separate a core library.

goto40:
If we introduce a (new, third) "textx" project, depending on "textx-core" and "textx-dev", we **would be backward compatible**!
 * textx, depends on
   * textx-core (depends on arpeggio)
   * textx-dev (depends on click and friends + textx-core)

Here, "normal" developers and users will just install "textx" (as before).

igordejanovic:
Yup, your point is valid. This is a recurrent idea. I had it in the past and didn't carry it out for the same worries you have. It is convenient to have textx command around but at the same time it is not quite justified to have it deployed to the production. The whole cli and export functionality is used during development.

The idea to have a kind of a wrapper project that reference core and dev parts is also not bad but that project/package would be practically empty, just would state the dependencies so why not just install `textX-dev` which has a dependency to textX.

Maybe to turn it around and say let's extract core functionality to textX-core and the current textX will be the development environment. Thus for a production one should only install textX-core if he wants to avoid dependency overhead.

Another change I'm thinking is how to make CI easier. We will now have separate git repos and the number will grow. I have already implemented scaffolding as one language project (questionnaire) and a generator project. From a organizational/architectural point of view this is really nice but it makes CI harder as each PR might depend on dev version of those other git repos. The question is how to make those in sync. One option is going mono-repo, but that might lead to other problems. Something to think about...

igordejanovic:
As a side note if would be nice to convert `render_all_grammars` to a proper textx command.

goto40:
> As a side note if would be nice to convert `render_all_grammars` to a proper textx command.

Yes. But the naming of the output is a little bit of a "hack". It is not clear to me, what a good way to name the outfiles is... Without the naming "problem" a simple "unix-find" would do the job (at least the generation of pu/dot files).

goto40:
I like your idea of having just a "textx-core" extracted (no explicit "dev")! Is this really backward incompatible? It is just an extra dependency...

And why not adding an extra library "textx-dev" with scaffolding, export etc...? This, in turn, in backward incompatible, because export.py is moved.

igordejanovic:
> Yes. But the naming of the output is a little bit of a "hack". It is not clear to me, what a good way to name the outfiles is... Without the naming "problem" a simple "unix-find" would do the job (at least the generation of pu/dot files).

I think it is fine as it is. You are constructing the name based on a original name and a relative path from the given start folder. That's perfectly fine. So you can make a textx subcommand (like e.g. `check`) which accepts a start folder and optionally output folder. You then do what you are already doing in a script.

> I like your idea of having just a "textx-core" extracted (no explicit "dev")! Is this really backward incompatible? It is just an extra dependency...

Yeah, it might be actually a fully compatible. I was thinking to use this opportunity to drop support for older Python versions and some other more drastic changes.

> And why not adding an extra library "textx-dev" with scaffolding, export etc...? This, in turn, in backward incompatible, because export.py is moved.

It seems that the textX project would be left empty which would look quite awkward. I haven't seen such python packages so far :)

goto40:
OK. Then we will go with textX + textx-core? Looks good for me... I could make the render-all-grammar command once this PR is merged...

goto40:
Once this pr is 'ready' (together with textX - core) I could try it out before merging...

goto40:
Why drop 2.7, if it works?
Dropping 3.0-3.5 is less critical in my eyes...

igordejanovic:
> OK. Then we will go with textX + textx-core? Looks good for me... I could make the render-all-grammar command once this PR is merged...

Let's sleep over and think some more. This is a big change so I wouldn't rush it. I think some of your concerns above are on point so I'll think a bit more about all pros and cons :)

> Why drop 2.7, if it works?

Python 2 is EOL so we should drop it sooner or later so we can move towards new features and syntax. We can leave it for some time more until the tools/libs we are using are supporting it.

igordejanovic:
Ok, here is a new suggestion:

1. Leave textX as is. Consider these few cli commands part of the core.
2. Additional commands for development go to `textX-dev` which is an optional `dev` dependency of textX. So for a full dev environment one has to install textX with `pip install textX[dev]`.
3. Move scaffolding generator from `textx-gen-que-txproject` to `textX-dev`. No need for this separation as the generator is not usable by itself.
4. Leave questionnaire language in `textx-lang-questionnaire` as this language is usable on its own. This project is a dependency of `textX-dev`.
5. Move `render_all_grammars` to `textX-dev` and make a proper `textx` subcommand out of this script.

I think these changes won't be disturbing at all. More or less everything will remain backward compatible while we will have `textX-dev` for dev-only stuff and dependencies. Thoughts?

igordejanovic:
@goto40 I've implemented changes as state above. It looks much better to me now. Changes are small and we are supporting Python 2.7 for both textX and textX-dev. Please review when you find time.

Next steps would be to:
- Improve project scaffolding template to the point where a fully functional language/generator project is generated when scaffolding is run.
- Better scaffolding tests (run -> install generated project -> verify it is registered)
- Document this new feature and make necessary changes/notes for this dev separation
- Move `render_all_grammars` to textX-dev and implement it as textx subcommand (not critical, can be done sometimes later)
- New release

goto40:
I will review the changes the next days

setup.cfg
```diff
@@ -1,5 +1,76 @@
...
+    textx = textx.cli:textx
+
+textx_commands = 
+    version = textx.cli.version:version
+    list_languages = textx.cli.discover:list_languages
+    list_generators = textx.cli.discover:list_generators
+    generate = textx.cli.generate:generate
+    check = textx.cli.check:check
+
+textx_generators =
```
goto40, line 67: Maybe also update the demo projects in tests/functional/registration/projects

CONTRIBUTING.md
```diff
@@ -82,15 +82,13 @@ Ready to contribute? Here's how to set up `textX` for local development.
...
+3. Install your local copy into a virtualenv. This is how you set up your fork
+   for local development:
 
         $ cd textX/
-        $ virtualenv venv
+        $ python -m venv venv
         $ source venv/bin/activate
-        $ pip install -r requirements_dev.txt
-        $ pip install -e .
+        $ ./install-dev.sh
```
goto40, line 91: This gave me
```
ERROR: Could not find a version that satisfies the requirement textX-dev (from textX==2.2.0.dev0) (from versions: none)   
```
(may be ok, because I test the branch)

goto40: I have a similar version issue while running `install-dev.sh` in textX-dev...

goto40: After install "https://github.com/textX/textx-lang-questionnaire" I managed to run the textx-dev tests successfully. OK!

goto40: After installing textX-dev manually, the script is working.

setup.cfg
```diff
@@ -1,5 +1,76 @@
...
+    click==7.0
+include_package_data = True
+package_dir =
+    textx = textx
+setup_requires =
+    wheel
+
+[options.extras_require]
+dev =
+    textX-dev
```
goto40, line 44: This is probably ok, but I had these errors above: Is it ok that "textX" (dev) depends on "textX-dev" here and "textX-dev" in turn depends on "textX" (a circular dependency?).

goto40 (approved):
Looks ok to me. Very good! Minor points:

- The new setup.py looks much better. **maybe also update the setup.py of our examples** in the test folder (to educate the unittest-reader) - see inline comments.

- I had these **install-dev.sh issues** (see inline comments), but if you think this is due to the intermediate state of the textX-dev project, just go ahead (I am not yet familiar with the "[dev]" suffix) - see inline comments.

CONTRIBUTING.md
igordejanovic: Yeah, textx-lang-questionnaire is not yet deployed to PyPI, so once we get that uploaded it will be pulled in from there. Also, textX-dev is not yet deployed, so once it does you will be able to do `pip install textX[dev]` to pull in optional `dev` dependencies. 

That `[dev]` suffix is a setuptools way to specify dependencies that are optional and only needed if some particular feature is desired. That extra dependencies can be called however you like and if you want to install some of those you just use `[]` suffix with the names of optional deps separated by commas. I named these dependencies `test` and `dev` for obvious reasons :) This should be cleaner than using separate requirements files.

setup.cfg
igordejanovic: Yup, good question. It should be ok. textX has textX-dev as a development dependency, so you need to pull in `dev` dependency if you want to use additional textX dev tooling (currently just `startproject` command but I expect it to grow over time). One way would be to install `textX-dev` and pull `textX` as a dependency but I thought that it might be nicer to install textX directly and if additional dev tooling is needed than you could use `textX[dev]` in the same way you could use `textX[test]` to pull in test dependencies. It seems more consistent to me, and the helper scripts seems to be more consistent across the  projects.

setup.cfg
igordejanovic: Yeah, good point. It would be good to have a consistency across all examples and tests.

goto40:
Thank you for all the explanation!

PR #233, opened 2020-02-17
Extract common logic for generators that produce a single file

igordejanovic:
@goto40 A common logic for simple generators is extracted and dot/puml exporter are reworked to use this.

## Code review checklist

### DIFF
--- a/textx/generators.py
+++ b/textx/generators.py
@@ -1,43 +1,72 @@
 import click
 import os
+from functools import partial
 from textx import generator
 from textx.export import metamodel_export, model_export, PlantUmlRenderer
+
+
+def get_output_filename(input_file, output_path, fileext):
+    """
+    Helper function used to create output file based on output path, input_file
+    base name and extension.
+    Args:
+        fileext (str): Target file name extension.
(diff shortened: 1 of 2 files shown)


textx/generators.py
```diff
@@ -1,58 +1,83 @@
...
-from textx import generator
+from functools import partial
+from textx.registration import generator
 from textx.export import metamodel_export, model_export, PlantUmlRenderer
 
 
-@generator('textX', 'dot')
-def metamodel_generate_dot(metamodel, model, output_path, overwrite, debug):
-    "Generating dot visualizations from textX grammars"
+def get_output_file(input_file, output_path, fileext):
```
goto40: I would suggest to call the function `get_output_filename` or `create_output_filename`, because it is not clear if a file is returned or a filename.

goto40 (approved):
Very good! This makes usage much simpler! Minor comment: see inline comment.

textx/generators.py
igordejanovic: @goto40 Good catch! Thanks.

PR #237, opened 2020-03-25
Click fixes

igordejanovic:
This PR:
- fixes the version spec for Click (relaxes it to use any version between 7.0 and 8.0)
- removes wrapping of bare `Exception` with `ClickException` as we shouldn't expect anything than textX exceptions in the normal flow of work. If any other exception is raised we should throw it unchanged.

## Code review checklist

[commit messages]: https://chris.beams.io/posts/git-commit/7.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,8 +31,8 @@ classifiers =
 packages = textx, textx.scoping, textx.cli
 zip_safe = False
 install_requires =
-    Arpeggio>=1.9.0
-    click==7.0
+    Arpeggio >= 1.9.0
+    click >=7.0, <8.0
 include_package_data = True
 package_dir =
     textx = textx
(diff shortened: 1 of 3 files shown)


goto40:
Looks good... I added the ClickException mainly because I had an issue with the exit code (not indicating an error when some problem occurred).

This seems to work now. However, I get a strange effect (although the unittests run successfully):

After running the unittests I get an error (`AttributeError: _tx_reference_resolver`) when doing this:
```
textx check tests/functional/registration/projects/flow_dsl/tests/models/data_flow.eflow
```
do you get the same? I do not understand what I messed up...

igordejanovic:
@goto40 I get the same error. It seems that `check` is broken for registered languages. Checked for released 2.1.0 and got the same error. Investigating.

igordejanovic:
@goto40 I figured it out. Root model was registered two times in `all_models.filename_to_model`, once with relative name and the other time with full path. I did a fix but erroneously pushed it to master. Please see the last commit on the master. If you agree with the fix we'll leave it there.

goto40:
Very good!

goto40 (approved):

PR #242, opened 2020-04-15
Allow passing project root path to GlobalModelRepository

danixeee:
Provides a way to search for models at different location regardless where python process is started (`os.getcwd`).

Related to #239 and #240.

Should we allow passing multiple locations?

- [x] `project_root` param when parsing models (`model_from_str`, `model_from_file`)
- [ ] `project_root` param when parsing meta-models (?)
- [ ] cli `project_root` param

## Code review checklist

- [ ] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [ ] Title summarizes what is changing
- [ ] Commit messages are meaningful (see [this][commit messages] for details)
- [ ] Tests have been included and/or updated
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly
- [ ] Changelog(s) has/have been updated, as needed (see `CHANGELOG.md`, no need
      to update for typo fixes and such).

[commit messages]: https://chris.beams.io/posts/git-commit/

Issue #240, opened 2020-04-13
Passing parameters to a metamodel (while opening a model) 


goto40:
While discussing #239 I realized that we may need a way to pass parameters to a metamodel (while loading a model). Such parameters could influence scope providers (searchpath options or recursive globbing) or model checkers (object processors checking certain rules).

The parameter capabilities should be queryable via the meta-model ('which parameters exist?'). With such, an appropriate command-line interface _and_ ide interface could be deduced.

We could also solve #239 with that. It allows to keep the metamodel stateless and to influence the meta-model behavior on a per-file basis. 

What do you think?

igordejanovic:
So, adding optional set of kw parameters to `model_from_file/str` of the meta-model. I like the idea. It would make it a kinda symmetric to `metamodel_from_file/str` and could indeed make model loading more flexible.

Do we need capabilities querying on the meta-model? How the meta-model creator would define those and how we could utilize them in a generic way? What would be the use-case?

goto40:
I'm not sure about possible concrete implementations.

But the use case is to provide 
 * a CLI for the parameters: e.g. `textx check --lang-arg1=value1 --root-path=$(pwd) myfile.lang` (with the language parameters `lang-arg1` and `root-path`)
 * a easy IDE integration (e.g. provide a dialog with the possible parameters)

Details to be discussed.
Querying could be a special function of the metamodel returning the description: The meta model designer will just set the description: e.g., `metamodel.set_parameter_descrition("paramname", "meaning")`

Everything should be light-weight. 
 * All parameters are strings? This makes it easier for an IDE/CLI integration...
 * Parameters like search-paths or root-path/project-path could have a special meaning for the IDE (by convention). 
 * If a parameter is not set, the metamodel can do whatever the language designer wants to do (e.g. use CWD).

goto40:
Maybe @danixeee has some good ideas. Let's wait for his comments as well!

danixeee:
This sounds like a good solution to me and it should be easy to integrate with language server.

When document (model) is opened in a VS Code, _textX-LS_ finds appropriate `LanguageDesc` based on _language id_ or _file extension_ and then instantiate metamodel with `lang_desc.metamodel()`, so it will be really straightforward to pass `project_root` or any additional arguments there.
In VS Code, all projects/models could be under the same _workspace root_, or under separated _workspace folders_ and I could pass that path wherever we decide to.

goto40:
@igordejanovic : what do you think about the querying capabilities?

[...]

PR #243, opened 2020-04-18
Feature/metamodel parameters issue240

goto40:
## Code review checklist

- (squash + and reformulate) Commit messages are meaningful (see [this][commit messages] for details)

Steps for implementation:

 - [x] Add kwargs feature to `model_form_str` and `model_from_file`.
 - [x] Adapt scope provider
   - [x] GlobalRepo: `project_root` (for @danixeee , see #242)
   - [ ] glob_args??
   - [ ] search_path / use_search_path??
 - [x] Add parameter/kwargs description
 - [x] Define what to check (only described parameters may be specified?; All parameters must be used?)
        --> only defined parameters (in the metamodel) may be specified during model loading.
        --> the ability to check if all specified parameters were used is available (but not automatically
              executed; you can this, e.g., in a model processor).

 - [ ] Add a nice and easy-to-use CLI interface (click)
 - [x] check details of callback in internal_model_from_file(...)

Details: see #240 
The implementation needs to be discussed (comcrete API and features).


textx/metamodel.py
```diff
@@ -547,32 +548,52 @@ def _current_namespace(self):
...
 
             for p in self._model_processors:
                 p(model, self)
         else:
-            model = self.internal_model_from_file(file_name, encoding, debug,
-                                                  model_str=model_str)
+            model = self.internal_model_from_file(
+                file_name, encoding, debug,
+                model_str=model_str,
+                pre_ref_resolution_callback=kwargs_callback)
```
goto40: Hmm. This is new: before this change, no callback was passed in that case. Unclear to me (but all tests pass).

goto40:
Hi @danixeee ,
I tried to add `project_root` to enable your use case. Maybe you can take a look...

This PR is in progress (see checklist and plan below checklist)

goto40:
Hi @igordejanovic and @danixeee ,

This is a first sketch: It allows to pass kwargs and to use them in one scope provider... You may have a look.

Special thing added: The class `ModelKwargs` is a special dict to track all parameter usage (which may allow to make checks if we have unused parameters.

No parameter descriptions/specifications added so far...

I will continue - if you have already comments concerning the "way" the kwargs are passed, feel free to comment...

goto40:
I am missing the automated unittests (travis)? @igordejanovic: Did we change something concerning the automated unittest execution? Or did our travis config change?

igordejanovic:
> I am missing the automated unittests (travis)? @igordejanovic: Did we change something concerning the automated unittest execution? Or did our travis config change?

Not that I know. I've just logged in to Travis and it seems that it got stuck, didn't see any textX repo in the beginning but gradually everything got back to normal it seems. It seems to work now.

danixeee:
Do you plan to add a new field on _Metamodel_ and _Model_ classes level which will contain a possible metamodel/model kwargs?
As you have mentioned in https://github.com/textX/textX/issues/240#issuecomment-612829940, we could then register all _builtin_ params in the following way `Metamodel.set_parameter_description("project_root", "...")`, and add custom ones either on class or instance level. Also, we could extend `set_parameter_description` method with `required` and `default` params.

goto40:
Yes. I will propose a description of possible parameters next. Then we can discuss more details... Not sure about the required flag.

goto40:
I see descriptions on metamodel level. Concrete values on model level.

textx/metamodel.py
goto40: Fixed: reverted change and preserved 100% backwards compatibility.

tests/functional/test_scoping/test_global_import_modules.py
```diff
@@ -151,3 +153,41 @@ def test_globalimports_basic_test_with_distributed_model():
...
+    # META MODEL DEF
+    #################################
+
+    my_meta_model = metamodel_from_file(
+        join(abspath(dirname(__file__)), 'interface_model2',
+             'Interface.tx'))
+    my_meta_model.register_scope_providers(
+        {"*.*": scoping_providers.FQNGlobalRepo(
+            join("model_a", "*.if"))})
+    my_meta_model._tx_model_kwarg_definitions.add_definition(
```
goto40, line 173: This is how a definition can be defined at the current state of implementation. At the moment there is no concept of automatically define parameters of, e.g., scope providers. You need to know that the GlobalRepo evaluates a parameter called "project_root".

However, you could check that the parameter was consumed by a concrete model.

goto40:
This is how a parameter  can be defined at the current state of implementation for a metamodel. At the moment there is no concept of automatically define parameters of, e.g., scope providers. As a metamodel designer, you need to know that, e.g., the GlobalRepo evaluates a parameter called "project_root":

```
my_meta_model._tx_model_kwarg_definitions.add_definition( "project_root", "the root of your project")
```

However, you could check that the parameter was consumed by a concrete model.

You can also query the possible parameters via `metamodel._tx_model_kwargs_definitions` (see model.md and unittest).

docs/model.md
```diff
@@ -190,3 +190,14 @@ The model may have a model repository (initiated by some scope provider or by
 the metamodel). This object is responsible to provide and cache other model
 instances (see textx.scoping.providers).
 
+### _tx_model_kwargs
```
igordejanovic: I would call this `_tx_model_params` as they are `kwargs` just in the API calls but they actually represent model parameters.

tests/functional/test_metamodel/test_model_kwargs.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+
+text = r"""
+MyModel test1
+"""
+
+
+def test_model_kwargs():
+    mm = metamodel_from_str(grammar)
+    mm._tx_model_kwarg_definitions.add_definition(
```
igordejanovic: This would thus be `_tx_model_param_defs`  and the `add_definition` can just be `add` it is implied what we are adding by the name of the container.

tests/functional/test_metamodel/test_model_kwargs.py
```diff
@@ -0,0 +1,94 @@
...
+        "parameter2", "an example param (2)"
+    )
+
+    m = mm.model_from_str(text, parameter1='P1', parameter2='P2')
+    assert m.name == 'test1'
+    assert hasattr(m, '_tx_model_kwargs')
+    assert len(m._tx_model_kwargs) == 2
+    assert len(m._tx_model_kwargs.used_keys) == 0
+
+    assert not m._tx_model_kwargs.have_all_parameters_been_used()
```
igordejanovic: `have_all_parameters_been_used` can be called just `all_been_used` and that could be Python property so no need for `()`

BTW, do we need checking if params have been used? We should make those params fully optional and the language designer should be aware that some/all of them might not be defined during model loading in which case a sane default should be used.

tests/functional/test_metamodel/test_model_kwargs.py
```diff
@@ -0,0 +1,94 @@
...
+
+    assert m._tx_model_kwargs['parameter2'] == 'P2'
+    assert len(m._tx_model_kwargs.used_keys) == 2
+    assert 'parameter1' in m._tx_model_kwargs.used_keys
+    assert 'parameter2' in m._tx_model_kwargs.used_keys
+
+    assert m._tx_model_kwargs.have_all_parameters_been_used()
+
+    assert m._tx_model_kwargs.get_with_default(
```
igordejanovic: `get_with_default` should be just `get` where `default` param is optional and by default `None` to make it similar to `dict`

igordejanovic (commented):
@goto40 I did just a light review. Looks good so far.

I have some suggestion, mostly regarding naming. Please see inline comments.

danixeee:
@goto40 Just to confirm that this works well with the language server, thanks!

goto40:
Thanks for the feedback. I need a few days...

docs/model.md
goto40: Good point. Done (wrong commit text; will be corrected while squashing).

tests/functional/test_metamodel/test_model_kwargs.py
goto40: done.

tests/functional/test_metamodel/test_model_kwargs.py
goto40: "have been used" --> "have been used by the metamodel" and not "have been specified by the user"...

I am not sure if we need this. It could be a hint, that you accidentally provided an argument not used by your meta model... 

What do you think... It is easy to remove this option. In that case we could also use a simple dict instead of the ModelParams class.

tests/functional/test_metamodel/test_model_kwargs.py
goto40: That was easy! I learned something very nice.

goto40: done.

tests/functional/test_metamodel/test_model_kwargs.py
goto40: (added property).

goto40: I think we should leave this option: if an option passed by the user has not been used, this is at least a warning in some cases... (maybe checked by a model processor).

goto40:
Maybe @igordejanovic, you want to have a deeper look.

If not:
 * I adapted the changes proposed by @igordejanovic .
 * I propose to leave the command line interface for a future PR. We have the playground defined in #245 as a base...
 * For now @danixeee can continue with our IDE support...

What do you think?

goto40:
Additional "default parameters" (search path, etc.) could be discussed in future.

However, one thing which may be of interest at the current state: do we plan to **"automatically" add some parameter definitions** if, e.g., a certain **scope provide is employed** or **another metamodel is referenced**...

textx/scoping/providers.py
```diff
@@ -462,9 +468,13 @@ def register_models(self, filename_pattern):
 
     def _load_referenced_models(self, model, encoding):
         for filename_pattern in self.filename_pattern_list:
+            if not isabs(filename_pattern) and \
```
danixeee, line 471: `filename_pattern = join(model._tx_model_params.get('project_root', os.getcwd()), filename_pattern)` should work even if `filename_pattern` is absolute path.

danixeee:
Does it make sense to register a `project_root` param in metamodel constructor automatically?

textx/scoping/providers.py
goto40: But if an absolute path is passed I think it should be left as it is... (my opinion). @igordejanovic what do you think?

goto40: But if an absolute path is passed I think it should be left as it is... (my opinion). @igordejanovic what do you think?

goto40: But if an absolute path is passed I think it should be left as it is... (my opinion). @igordejanovic what do you think?

goto40:
We could define a base class 'autoregister' to allow scope providers to register themselves (maybe while passing it to the metamodel)...

textx/scoping/providers.py
danixeee: From [docs](https://docs.python.org/3/library/os.path.html?highlight=os%20path%20join#os.path.join): 
> ...
If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.
...

`join('/a/b/c', '/d/e/f')` returns `/d/e/f`.

danixeee:
I am not sure about the implementation, but I thought if we are already trying to get "project_root" somewhere in textX source code, maybe it would make sense to have that parameter predefined for every user and for CLI.

textx/scoping/providers.py
goto40: OK. You say, I do not need the 'if', because `join` already does the job... Good point. But the behavior will remain as it is...

igordejanovic:
@goto40 I'm running another round of review. Will take some time.

BTW, fully agree. All model parameters used in the textX source code should be implicitly defined on the meta-model. For example, we should not have to do [this](https://github.com/textX/textX/pull/243/files#diff-98b48d145b5db17325d4d3c881b08a7eR173), `project_root` should be available on every meta-model as that parameter is used by the textX source code.

@goto40 also agree on the CLI support. That can be dealt with another PR

docs/model.md
```diff
@@ -190,3 +190,14 @@ The model may have a model repository (initiated by some scope provider or by
...
+
+This attribute always exists. It holds all additional parameters passed to
+`model_from_str` or `model_from_file` of a metamodel. These parameters are
+restricted by the `metamodel._tx_model_param_definitions` object, which is
+controlled by the metamodel designer. 
+
+`metamodel._tx_model_param_definitions` can be queried (like a dict) to
+retrieve possible extra parameters and their description for a metamodel.
+It is also used to restrict the additional parameters passed to
+`model_from_str` or `model_from_file`.
```
igordejanovic: We should extend this section (or maybe add something in the metamodel docs) to specify how new params are defined and used in obj. processors for example.

tests/functional/test_scoping/test_global_import_modules.py
```diff
@@ -151,3 +153,41 @@ def test_globalimports_basic_test_with_distributed_model():
...
+    #################################
+
+    my_meta_model = metamodel_from_file(
+        join(abspath(dirname(__file__)), 'interface_model2',
+             'Interface.tx'))
+    my_meta_model.register_scope_providers(
+        {"*.*": scoping_providers.FQNGlobalRepo(
+            join("model_a", "*.if"))})
+    my_meta_model._tx_model_param_definitions.add(
+        "project_root", "the project root")
```
igordejanovic: `project root` is a built-in param so language designer shouldn't need to define it explicitely.

textx/metamodel.py
```diff
@@ -164,6 +165,7 @@ def __init__(self, file_name=None, classes=None, builtins=None,
 
         super(TextXMetaModel, self).__init__(**kwargs)
 
+        self._tx_model_param_definitions = ModelParamDefinitions()
```
igordejanovic, line 168: This looks like a good place to introduce textX built-in model param definitions.

textx/model_params.py
```diff
@@ -0,0 +1,98 @@
...
+        return key
+
+    def _have_all_parameters_been_used(self):
+        return reduce(
+            lambda r, k: r and (k in self.used_keys),
+            self.store.keys(), True)
+
+    all_been_used = property(
+        lambda self: self._have_all_parameters_been_used(),
+        doc="returns if all parameters have been used by the meta model")
```
igordejanovic: This part of code (property def and previous method) can be replaced with:
```python
        @property
        def all_been_used(self):
            "returns if all parameters have been used by the meta model"
            return not (set(self.store.keys()) - set(self.used_keys))
```

Another shortening suggestion is making this property called `all_used`.  It is shorter to read/type while not being less understandable

igordejanovic (commented):
@goto40 Good work. I have some suggestions. Please see bellow.

textx/model_params.py
goto40: done.

tests/functional/test_scoping/test_global_import_modules.py
goto40: ok

textx/metamodel.py
goto40: ok. also adapated tests

docs/model.md
goto40: Done.

goto40: Added an example and a unittest.

goto40:
Thank you for the comments. I added some documentation. Maybe you have time to check it...

docs/metamodel.md
```diff
@@ -318,3 +318,27 @@ assignments (`=`). An optional assignment (`?=`) will always be `False` if the
 RHS object is not matched in the input. The multiplicity assignments (`*=` and
 `+=`) will always be python lists.
 
+## Optional model parameter definitions
+
+A metamodel can define optional model parameter definitions. Such definitions
```
igordejanovic: A meta-model can define optional model parameters. Such definitions...

docs/metamodel.md
```diff
@@ -318,3 +318,27 @@ assignments (`=`). An optional assignment (`?=`) will always be `False` if the
...
+are stored in `_tx_model_param_definitions` and define optional parameters,
+which can be specified while loading/creating a model through `model_from_str`
+or `model_from_file`. Details: see [tx_model_params](model.md#_tx_model_params).
+
+`metamodel._tx_model_param_definitions` can be queried (like a dict) to
+retrieve possible extra parameters and their description for a metamodel.
```
igordejanovic: their descriptions
meta-model

docs/metamodel.md
```diff
@@ -318,3 +318,27 @@ assignments (`=`). An optional assignment (`?=`) will always be `False` if the
...
+It is also used to restrict the additional parameters passed to
+`model_from_str` or `model_from_file`.
+
+Default parameters are:
+
+ * `project_root`: This parameter defines the root of the project and is used by
```
igordejanovic: ...defines the root path of the project...

docs/model.md
```diff
@@ -190,3 +190,12 @@ The model may have a model repository (initiated by some scope provider or by
...
 
+### _tx_model_params
+
+This attribute always exists. It holds all additional parameters passed to
+`model_from_str` or `model_from_file` of a metamodel. These parameters are
+restricted by the `metamodel._tx_model_param_definitions` object, which is
+controlled by the metamodel designer.
+
+See also:
+[model and object processors](metamodel.md#optional-model-parameter-definitions).
```
igordejanovic: This could be a removed and the link placed on mentioning `metamodel._tx_model_param_definitions` object above.

textx/metamodel.py
```diff
@@ -164,6 +165,10 @@ def __init__(self, file_name=None, classes=None, builtins=None,
...
+        self._tx_model_param_definitions.add(
+            "project_root", "the project root")
```
igordejanovic: the project root path

textx/metamodel.py
```diff
@@ -547,43 +552,67 @@ def _current_namespace(self):
...
         Instantiates model from the given string.
         :param pre_ref_resolution_callback: called before references are
                resolved. This can be useful to manage models distributed
                across files (scoping)
+        :param **kwargs additional arguments available through
+                _tx_model_params after initiating the model. The attribute
+                is set while executing pre_ref_resolution_callback (see
+                scoping.md)
         """
+        self._tx_model_param_definitions.check_params_and_raise_on_error(
```
igordejanovic: Sorry, looks like I forgot to comment on this in the last review. `check_params_and_raise_on_error` should just be `check_params`. Much simpler to read/write, no need to encode behavior in the name.

textx/scoping/providers.py
```diff
@@ -439,6 +441,10 @@ class GlobalRepo(ImportURI):
       * create FQNGlobalRepo
       * register models used for lookup into the scope provider
     Then the scope provider is ready to be registered and used.
+
+    The model parameter `project_root` (see _tx_model_params) can be used to
+    set a project directory, where all file patterns not referring to an
+    absolute file position are looked up.
```
igordejanovic, line 447: This explanation looks good, so might go to the docs for explanation of the parameter.

igordejanovic (commented):
@goto40 Thanks. Looks better. There are few more suggestions/fixes. Please see below.

docs/metamodel.md
goto40: ok

docs/metamodel.md
goto40: done

docs/metamodel.md
goto40: done

docs/model.md
goto40: done

textx/metamodel.py
goto40: ok

textx/metamodel.py
goto40: no problem. if you find more stuff - comments welcome!

goto40: done (2x)

textx/scoping/providers.py
goto40: Done. Retsricted to globalrepo. Also added a link to scoping.md (for the globalrepo)

igordejanovic:
@goto40 LGTM. This is ready for squashing and merging. Maybe @danixeee could give it another spin with textX-LS just to be sure?

danixeee:
@igordejanovic @goto40 Tested it again, works well.

igordejanovic:
@danixeee Thanks.
@goto40 Great work, no more addition to this, feel free to squash merge this PR.

PR #245, opened 2020-04-19
click experiment: demo for autocompletion

goto40:
This PR represents an experiment how to integrate click completition (for the kwargs PR). I extended the types-dsl language to have an extra kwargs option.

**Preparation**
Run in your virtualenv (I use a bash)
```
_TEXTX_COMPLETE=source_bash textx > ./venv/bin/textx-complete.sh
. ./venv/bin/textx-complete.sh
```

**Demo**
Then, you can do `textx <TAB> <TAB>` to get cool autocompletions... (it allows to select a language for `textx check --language ...` and also one optional kwargs value):
```
textx check --language types-dsl --x <TAB> <TAB>
```

Of course we will also provide some "list-all-possible-options..."...

Open for discussion!

## Code review checklist


igordejanovic:
@goto40 This looks cool. Great idea!
I see that autocomplete should provide all options by pressing <TAB> <TAB> after a dash. Maybe we can integrate into this to add dynamically model params as well if the language is specified, e.g. either by `--language` or by giving the model file name with extension?

goto40:
Autocompletition works great for options integrated with decorators (as we do in textX). E.g.,--language automatically provides all available languages (this feels great). 

However I did not manage, so far, to _dynamically_ add an option (based on the language). Maybe I have to investigate further.

So far, I provided an option to specify a param: - - x paramname=value
I do not like it very much - it feels like a workaround... (but it works).

goto40:
The - - x option autocompletes, based on the selected language.

igordejanovic:
@goto40 I've been playing today with this feature. One observation is that while we get autocompletion on textX languages and options, which is great, we loose completion on file names (e.g. calling `textx check ` and then trying to complete on a file name). Didn't find anything in the click docs how to keep this normal behavior for arguments and options which doesn't define custom completer. One way we could do it is to implement completer completing on file names and applying it on all arguments which should be files. That could be a nice solution as file name completion would happen only where it make sense.

BTW, this is probably not a problem of Click but the way shell completion works.

igordejanovic:
I would really like to be able to filter/extend list of options for a particular command. This would enable to extend the list if a language is given thus eliminating the need to remember that you need to type `--x` to complete on language params. It seems that it is not possible with the current version of Click though. We might investigate if it is possible to propose an extension of this Click feature.

goto40:
> I would really like to be able to filter/extend list of options for a particular command. This would enable to extend the list if a language is given thus eliminating the need to remember that you need to type `--x` to complete on language params. It seems that it is not possible with the current version of Click though. We might investigate if it is possible to propose an extension of this Click feature.

So you would like to dynamically add `--root-path ~/folder` (**nice**) instead of `--x "root_path=~/folder"` (**ugly**). I did not manage to do this in the first version proposed here, be we should investigate. I fully agree.

I will further try to play with click.. if you find something you can comment here.

goto40:
(I accidentally did some refactorings in this branch - this should not hurt - this is a playground only)

PR #248, opened 2020-04-29
[WIP] Store attributes of user classes in separate dictionary

markusschmaus:
## Code review checklist

- [x] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [x] Title summarizes what is changing
- [x] Commit messages are meaningful (see [this][commit messages] for details)
- [x] Tests have been included and/or updated
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly
- [ ] Changelog(s) has/have been updated, as needed (see `CHANGELOG.md`, no need
      to update for typo fixes and such).

[commit messages]: https://chris.beams.io/posts/git-commit/

Issue #247, opened 2020-04-29
Incompatible with attrs frozen class


markusschmaus:
I'm using attrs frozen classes to represent data with correct equal and hash semantics. Unfortunately these classes are incompatible with textX.

```python
import attr
from textx import metamodel_from_str

grammar = """
Point: x=INT ',' y=INT;
"""

@attr.s(frozen=True)
class Point:
    x = attr.ib()
    y = attr.ib()

point1 = Point(0, 0)
mm = metamodel_from_str(grammar, classes=[Point])
point2 = mm.model_from_str("0, 0")
```

This is also true for classes which use `__slots__`

```python
class Point:
    __slots__ = ['x', 'y']

    def __init__(self, x, y):
        self.x = x
        self.y = y
```

markusschmaus:
I created an implementation which stores the attributes of user classes in a separate dictionary:

https://github.com/markusschmaus/textX/commit/3a9aca2eb6b4656fa5342990784c425e549f47f7

All tests that functional tests that succeed on my system for master also succeed on this branch (33 tests also fail for master) and I added a small test for a user class that uses `__slots__`.

igordejanovic:
Thanks for reporting and working on it. Looks good so far. I wonder for the failing tests you are getting. All functional tests should pass for all supported Python versions. What version of Python are you using? How is your environment setup? Can you tell from the error reports you are getting what would be the reason?

Please register a WIP PR for your working branch, I think this issue is important and would like to get it merged once it is finished.

igordejanovic:
@markusschmaus I'm confirming that all tests are passing on  my end with your changes.

markusschmaus:
https://github.com/textX/textX/pull/248

Great to hear that everything is passing on your end, I think the problem was with finding files, probably because I'm in the wrong working directory.

markusschmaus:
My original PR didn't fully address the issue, https://github.com/textX/textX/pull/249

PR #249, opened 2020-04-30
Store attributes of user classes outside object and postpone initiali…

markusschmaus:
…zation

## TODOS
- [ ] Change the merge target branch of this PR to next-release (@markusschmaus)
- [ ] Change _has/get/setattr to s_has/get/setattr (@markusschmaus)
- [ ] Update scoping docs to incorporate new changes (@goto40)
- [x] Update user class docs to mention __slots__ and immutable objects (@igordejanovic #252)
- [x] Check how performance is affected by this change (@igordejanovic)

## Code review checklist

This PR creates compatibility with attrs frozen classes (https://github.com/textX/textX/issues/247).
Unlike https://github.com/textX/textX/pull/248 this also allows linked references for such classes.

To achieve compatibility two changes were necessary. First the attributes which are passed to `__init__` are stored outside the object in a dictionary of the class. Second, initialization of the object is postponed until all references have been resolved.

The first change also allows the user to use `__slots__` classes (without using attrs).

Postponing initialization is necessary since the attributes of attrs frozen classes cannot change after initialization, this includes references to other classes.

One tricky aspect of this change is to make sure that as long as user class objects are not initialized any attributes like `name` and `parent` are only accessed via the `_getattr` helper functions. In order to find all such instances I searched for any instance of `hasattr` which checks for the presence of such an attribute.

@goto40, @igordejanovic please review this PR

### DIFF
--- a/tests/functional/test_processors.py
+++ b/tests/functional/test_processors.py
@@ -92,6 +92,54 @@ def test_object_processors():
     assert call_order == [2, 2, 2, 1]
 
 
+def test_object_processors_user_classes():
+    """
+    Test that object processors are called.
+    They should be called after each model object construction.
+    """
+
+    def first_obj_processor(first):
+        first._first_called = True
+        first._a_copy = first.a
+
+    def second_obj_processor(second):
+        second._second_called = True
+        second._sec_copy = second.sec
+
+        # test that parent is fully initialised.
+        # b should be True
+        assert second.parent.b is not None
+
+    obj_processors = {
+        'First': first_obj_processor,
+        'Second': second_obj_processor,
+        }
+
+    class First(object):
+        def __init__(self, seconds, a, b, c):
+            self.seconds = seconds
+            self.a = a
+            self.b = b
+            self.c = c
+
+    class Second(object):
+        def __init__(self, sec, parent):
+            self.sec = sec
+            self.parent = parent
+
+    metamodel = metamodel_from_str(grammar, classes=[First, Second])
+    metamodel.register_obj_processors(obj_processors)
+
+    model_str = 'first 34 45 7 A 45 65 B true C "dfdf"'
+    first = metamodel.model_from_str(model_str)
+
+    assert hasattr(first, '_first_called')
+    assert first._a_copy == first.a
+    for s in first.seconds:
+        assert hasattr(s, '_second_called')
+        assert s._sec_copy == s.sec
+
+
 def test_object_processor_replace_object():
     """
     Test that what is returned from object processor is value used in the
--- a/tests/functional/test_user_class_attrs_frozen.py
+++ b/tests/functional/test_user_class_attrs_frozen.py
@@ -1,0 +1,51 @@
+from __future__ import unicode_literals
+import pytest  # noqa
+from textx import metamodel_from_str
+
+grammar = """
+Document:
+    a=A
+    b=B
+;
+A:
+    'A' name=ID
+;
+B:
+    'B' 'a' '=' a=[A]
+;
+"""
+
+
+@pytest.mark.parametrize('frozen', [False, True])
+def test_user_class_attrs(frozen):
+    attr = pytest.importorskip('attr')
+    """
+    User supplied meta class.
+    """
+    @attr.s(frozen=frozen)
+    class Point(object):
+        "User class."
+        parent = attr.ib()
+        name = attr.ib()
+        x = attr.ib()
+        y = attr.ib()
+
+    modelstr = """
+    A something
+    B a=something
+    """
+
+    @attr.s()
+    class A(object):
+        parent = attr.ib()
+        name = attr.ib()
+
+    @attr.s(frozen=frozen)
+    class B(object):
+        parent = attr.ib()
+        a = attr.ib()
+
+    mm = metamodel_from_str(grammar, classes=[A, B],
+                            auto_init_attributes=False)
+    model = mm.model_from_str(modelstr)
+    assert model.b.a == model.a
--- a/tests/functional/test_user_class_with_slots.py
+++ b/tests/functional/test_user_class_with_slots.py
@@ -1,0 +1,48 @@
+from __future__ import unicode_literals
+import pytest  # noqa
+from textx import metamodel_from_str
+
+grammar = """
+Shape:
+    'shape'
+    points+=Point
+    'end'
+;
+Point:
+    'point' x=INT y=INT
+;
+"""
+
+
+def test_user_class_with_slots():
+    """
+    User supplied meta class.
+    """
+    class Point(object):
+        "User class."
+        __slots__ = ['parent', 'x', 'y']
+
+        def __init__(self, parent, x, y):
+            self.parent = parent
+            self.x = x
+            self.y = y
+
+    modelstr = """
+    shape
+    point 34 45
+    point 12 23
+    end
+    """
+
+    mm = metamodel_from_str(grammar, classes=[Point],
+                            auto_init_attributes=False)
+    model = mm.model_from_str(modelstr)
+    # Test that user class is instantiated
+    point = model.points[0]
+
+    assert type(point).__name__ == "Point"
+    assert type(point) is Point
+
+    assert len(model.points) == 2
+    assert point.x == 34
+    assert point.y == 45
--- a/tests/functional/test_user_classes.py
+++ b/tests/functional/test_user_classes.py
@@ -11,7 +11,7 @@ First:
 ;
 
 Second:
-    INT|STRING
+    sec=INT|STRING
 ;
 
 """
@@ -21,6 +21,7 @@ def test_user_class():
     """
     User supplied meta class.
     """
+
     class First(object):
         "User class."
         def __init__(self, seconds, a, b, c):
@@ -33,6 +34,16 @@ def test_user_class():
             self.b = b
             self.c = c
 
+            for second in self.seconds:
+                # Make sure seconds have already been instantiated
+                assert hasattr(second, 'sec') and isinstance(second.sec, int)
+
+    class Second(object):
+        "User class"
+        def __init__(self, parent, sec):
+            self.parent = parent
+            self.sec = sec
+
     modelstr = """
     first 34 45 65 "sdf" 45
     """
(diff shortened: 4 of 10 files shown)


goto40:
Wow. Thank you very much for that work, @markusschmaus. Sounds tricky... I will have a deeper look, especially concerning scope providers...

markusschmaus:
I noticed that my change conflicted with object processors and I wrote a corresponding test and moved initialization before the processing of object processors.

markusschmaus:
I was using `get_children` to find all objects that need to be initiated, since it returns the parent ahead of the children, this meant that the parameters passed to the parent were still not processed yet and therefore in an uninitiated state. The simple fix i came up with was to iterate over the returned children in reversed order. This means that that last child will be initiated first.

goto40:
Hm...
I see that we have the following stages (compare http://textx.github.io/textX/stable/scoping/#using-the-scope-provider-to-modify-a-model), implemented in model.py
 1. **initial raw model w/o references** `model = process_node(parse_tree)`
     * Here, we create user objects `attrs = metamodel.obj_attrs.pop(id(obj_attrs))` and `inst.__init__(**attrs)` (**code A**) - not 100% sure that I got this correctly....
     * Note: object processors for match rules (e.g. regular expressions or rules like 'ID') are called somewhere here, bottom up during model construction. 
 2. **prepare model for full creation** (create a reference resolver) `_start_model_construction(model)`
 3. **load other model files** `scope_provider.load_models(model, encoding=encoding)` (repeat step 1-3, w/o steps 4-6)
 4. **resolve references if all models have been loaded** `if is_main_model: ...`
     * Here we set references/'pointers' to (existing) objects as textx-references into other object attributes `m._tx_reference_resolver.resolve_one_step()`-->`metamodel.setattr(obj, attr.name, resolved)`
     * Note, that the object to be resolved already exists at this point... (**code B**) 
 5. **terminate model creation** `_end_model_construction(m)` (cleanup resolver) for all loaded models
 6. **call obj processors** `call_obj_processors(m._tx_metamodel, m)` for all loaded models

I see a conflict in **code A** and **code B**, if the object is frozen (read-only):
  * The object is created in **code A** and will then be immutable
  * resolving the reference in **code B** violates immutability.

If we move the object creation to a later point (e.g. into step 5):
  * We could solve the problem,
  * However, we have to manage that we create references to proxies (**objects of resolved references do not yet exist at the point of scope resolution...**)

The minor problem of adapting the standard scope providers is less critical (we could make `textx_isinstance` capable of handling the proxies correctly). The main issue for me is the management of proxies and *references* to proxies (which have to be *corrected* when replacing the proxies with the concrete class). 

Possibly, the newly introduced `obj_attrs` already solve the thing completely - and have just to be executed in another part of the code?

@markusschmaus, @igordejanovic 
What do you think?
 * The proxies could be a nice enhancement... But we should carefully plan the implementation.
 * Or do we restrict the use of such immutable objects not to contain references?
 * Or any other ideas?
 * Or did I get the idea wrong... (compare the testfile: https://github.com/goto40/textx-playground/blob/master/tests/functional/test_scoping/test_metamodel_provider3_custom_classes.py#L11)

goto40:
maybe I missed an update (moved initialization before object processors)...I need to check

goto40:
@markusschmaus @igordejanovic Very good work!
A first test look promising now! I missed your last commits...

Next: I will have a look at the scope providers to make use of the obj_attrs...

goto40:
Sorry: again I somehow commented on #250

goto40:
(please check there - igor already commented as well)

igordejanovic:
@markusschmaus Great work Markus! @goto40 Thanks for the analysis.

So, this change affects resolving phase so scoping providers have to be careful to call modified version of has/get/setattr. Since users can write their own scoping providers this is backward incompatible change so we should merge to `next-release` branch (please see #176) to be published in v3.0. Also, since these modified version of `has/get/setattr` are public API (e.g. should be used by custom scoping providers) their names should not start with underscore. I suggest something like `s_hasattr` for `scoping hasattr`.

Before we get this merged I would also like to do some performance tests on larger models with user classes to check how the final search and initialization perform.

So here is the list of changes need to be done before this is merged (I put our names for each task but if you can't/don't want to do something feel free to remove yourself):

- [ ] Change the merge target branch of this PR to `next-release` (@markusschmaus)
- [ ] Change `_has/get/setattr` to `s_has/get/setattr` (@markusschmaus)
- [ ] Update scoping docs to incorporate new changes (@goto40)
- [ ] Update user class docs to mention `__slots__` and immutable objects (@igordejanovic)
- [ ] Check how performance is affected by this change (@igordejanovic)

@markusschmaus Please copy this list of task to the description of this PR at the top so it is easier for tracking.

markusschmaus:
@igordejanovic 
Which option would you prefer for changing the target?
1. merge next-release into this branch
2. rebase to next-release and force push
3. rebase to next-release in new branch and new PR
4. some other way

igordejanovic:
@markusschmaus Option 2. After rebase please squash all your commits into a single commit. Also please wait until [this](https://github.com/textX/textX/pull/250#issuecomment-623151986) is taken care of and incorporate that change also.

I asked a heavy user of user classes to test with their model (which are the biggest textX models I know of) and they are reporting that their tests are failing with this change. We are investigating the issue.

markusschmaus:
I have changed it to use a set of `collected_ids`

igordejanovic:
> I asked a heavy user of user classes to test with their model (which are the biggest textX models I know of) and they are reporting that their tests are failing with this change. We are investigating the issue.

It was a false alarm. Everything seems to be working as before. There are no noticable performance degradation. So we are all good.

goto40:
@markusschmaus I prepared some tests (https://github.com/goto40/textX/tree/msch_attrs_frozen) and I created the PR to your branch. Here, we can discuss (see below; next comment).

There, I will also add the docu mentioned by @igordejanovic ...

goto40:
@igordejanovic I have a strange effect unclear to me...

When using a custom class for the root model element (class Model), I break my test:

https://github.com/goto40/textX/blob/msch_attrs_frozen/tests/functional/test_scoping/test_scoping_tools.py#L64

maybe the specified scope provider is not selected in that case (and the default provider is utilized)... I will investigate the next days (maybe you have an idea).

igordejanovic:
@markusschmaus @goto40 For an easier sync I've created a branch `feature/postponed-userobj-init` which should collect all our changes for this feature. This branch is not used to directly work on it but to issue PRs against it. I've already created a PR for docs update #252 .

So, what we do is:
- Target our PRs against the feature branch. When we are done we'll squash merge to this branch.
- @goto40 Since you are working on tests you can keep you branch constantly rebased on top of this @markusschmaus branch.

At the end we should have only three commits on the feature branch which will be merged to `next-version` and released with the upcoming 3.x.

igordejanovic:
> @igordejanovic I have a strange effect unclear to me...

@goto40 I did a debugging session and everything seems to be right. The exception is correctly produced and propagated up the stack but strangely it seem not to be detected by `with raises` block. You can verify that the exception is produced by commenting out `with raises` line, the test will break with exception for the incorrect model. I am puzzled.

goto40:
> > @igordejanovic I have a strange effect unclear to me...
> 
> @goto40 I did a debugging session and everything seems to be right. The exception is correctly produced and propagated up the stack but strangely it seem not to be detected by `with raises` block. You can verify that the exception is produced by commenting out `with raises` line, the test will break with exception for the incorrect model. I am puzzled.

Thank you for the analysis: I tried to create a workaround and commented the effect in the test. --  However in my case I do not see the exception (only in the second pass of the loop - with custom classes). -- still failing... :-(

goto40:
I would like to encourage a **new discussion** about **how to activate the postponed initialization of custom objects**:

 * **At the moment** we create a postponed initialization for **all** `classes` (custom user classes) passed to the meta model.
 * This enables the language designer to use, e.g., immutable objects from libraries like `attr` and `dataclasses`.  
 * Our built-in scope providers can (or will can - if we find incompatibilities) cope with such custom classes (--> no problem).
 * Custom scope providers however, tend to get very complicated and **may irritate developers** - especially if they are not aware of the underlying design desicions (depicted in this PR):
**compare** `reference.refs[index - 1].valref` **vs.** `_getattr(_getattr(reference, "refs")[index - 1], "valref")`

Therefore I would like to propose to think about an alternative *activation of this feature*: We could **add an additional parameter to the metamodel**
  * Instead of `classes`
  * We could have `classes` and `classes_with_postponed_initialization` (we may find a better name). `classes` specifies **user classes (initialization as before)** and the other parameter specifies **user classes with postponed initialization**.

@igordejanovic @markusschmaus What do you think?

BTW : this PR gets backward compatible with this change...

igordejanovic:
> Custom scope providers however, tend to get very complicated and may irritate developers - especially if they are not aware of the underlying design desicions (depicted in this PR):
compare reference.refs[index - 1].valref vs. _getattr(_getattr(reference, "refs")[index - 1], "valref")

Yes, this change is the API was bugging me also. It doesn't look pretty. But, I would not like to introduce two different ways of working with user classes. While this will work as before for mutable classes it introduce complexity in the code-base and still users will be tripped when they do start using immutable classes. I was thinking about replacing `__*attr__` for user-classes during loading/resolution and restoring them back before call to obj. processor (in a `finally` block). It is a bit of hackery and monkey patching, and could bring problems down the road (maybe it is not doable for frozen classes, needs investigation), but looks like a good compromise at the moment favoring ease of use and backward compatibility.

markusschmaus:
> 
> 
> @markusschmaus @goto40 For an easier sync I've created a branch `feature/postponed-userobj-init` which should collect all our changes for this feature. This branch is not used to directly work on it but to issue PRs against it. I've already created a PR for docs update #252 .
> 
> So, what we do is:
> 
>     * Target our PRs against the feature branch. When we are done we'll squash merge to this branch.
> 
>     * @goto40 Since you are working on tests you can keep you branch constantly rebased on top of this @markusschmaus branch.
> 
> 
> At the end we should have only three commits on the feature branch which will be merged to `next-version` and released with the upcoming 3.x.

I guess this means I should retarget this PR to this feature branch. The three commits are probably three PRs. Are these the three PRs? one of them is this PR retargetted, one is https://github.com/textX/textX/pull/252, and one is a PR with @goto40 's tests?

igordejanovic:
@markusschmaus Yup, exactly.

markusschmaus:
retargetted

igordejanovic:
@markusschmaus Thanks for your work. 
I think we can merge this PR and for any further work we can open new PRs against the feature branch.
Would you like to squash your commits with interactive rebase or should I squash merge this PR?

markusschmaus:
Squash merge sounds good to me. BTW, I'm looking into some proxy based API for which I will create a separate PR

PR #250, opened 2020-04-30
Explicitly check for object to be not None rather than rely on truthi…

markusschmaus:
…ness

## Code review checklist

I found a few places in the code where an if condition relies on the truthiness of the corresponding object, this may lead to bugs when a user class object is falsy, most likely because it defines `__len__` to be 0. The test case I added tests one of these instances.

### DIFF
--- a/tests/functional/test_processors.py
+++ b/tests/functional/test_processors.py
@@ -92,6 +92,37 @@ def test_object_processors():
     assert call_order == [2, 2, 2, 1]
 
 
+def test_object_processor_falsy():
+    """
+    Test that object processors are called for falsy objects.
+    """
+
+    def second_obj_processor(second):
+        second._second_called = True
+
+    obj_processors = {
+        'Second': second_obj_processor,
+        }
+
+    class Second(object):
+        def __init__(self, parent, sec):
+            self.parent = parent
+            self.sec = sec
+            self._second_called = False
+
+        def __len__(self):
+            return 0
+
+    metamodel = metamodel_from_str(grammar, classes=[Second])
+    metamodel.register_obj_processors(obj_processors)
+
+    model_str = 'first 34 45 7 A 45 65 B true C "dfdf"'
+    first = metamodel.model_from_str(model_str)
+
+    for s in first.seconds:
+        assert s._second_called == True
+
+
 def test_object_processor_replace_object():
     """
     Test that what is returned from object processor is value used in the
--- a/tests/functional/test_processors.py
+++ b/tests/functional/test_processors.py
@@ -120,7 +120,7 @@ def test_object_processor_falsy():
     first = metamodel.model_from_str(model_str)
 
     for s in first.seconds:
-        assert s._second_called == True
+        assert s._second_called is True
 
 
 def test_object_processor_replace_object():
--- a/textx/model.py
+++ b/textx/model.py
@@ -510,7 +510,7 @@ def parse_tree_to_objgraph(parser, parse_tree, file_name=None,
                 assert False
 
         # Collect rules for textx-tools
-        if inst and metamodel.textx_tools_support:
+        if inst is not None and metamodel.textx_tools_support:
             pos = (inst._tx_position, inst._tx_position_end)
             pos_rule_dict[pos] = inst
 
@@ -553,7 +553,7 @@ def parse_tree_to_objgraph(parser, parse_tree, file_name=None,
                     if attr:
                         if metaattr.mult in many:
                             for idx, obj in enumerate(attr):
-                                if obj:
+                                if obj is not None:
                                     result = call_obj_processors(metamodel,
                                                                  obj,
                                                                  metaattr.cls)
--- a/textx/model.py
+++ b/textx/model.py
@@ -816,7 +816,7 @@ class ReferenceResolver:
                     resolved = default_scope(obj, attr, crossref)
 
                 # Collect cross-references for textx-tools
-                if resolved and not type(resolved) is Postponed:
+                if resolved is not None and not type(resolved) is Postponed:
                     if metamodel.textx_tools_support:
                         self.pos_crossref_list.append(
                             RefRulePosition(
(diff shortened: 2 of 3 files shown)


goto40:
I played with your code  yesterday and had some problems (not yet analyzed). I can check tonight or tomorrow - if you are intetrested: https://github.com/goto40/textx-playground

igordejanovic:
@markusschmaus LGTM. I found another case of skipping falsy objects with containment link. Added a fix and a test [to this branch based of yours](https://github.com/textX/textX/tree/msch_check_explicit_none) so you can pull it in to this PR. Also, fixed a comparison to truth value in assert that lead to tests failure on CI.

markusschmaus:
@igordejanovic 
https://github.com/markusschmaus/textX/blob/b91c7412a75fdb0b539f022e5ea55076c4f1bb97/textx/scoping/providers.py#L107 in combination with L110 converts a falsy result to `None`. Do you know if this is intentional?

igordejanovic:
@markusschmaus I believe it is not intentional but maybe @goto40 could take a look.

goto40:
> @igordejanovic
> https://github.com/markusschmaus/textX/blob/b91c7412a75fdb0b539f022e5ea55076c4f1bb97/textx/scoping/providers.py#L107 in combination with L110 converts a falsy result to `None`. Do you know if this is intentional?

@markusschmaus: It is not intentional. You can just `return result` instead of `if result: return result`and `return None`.

goto40:
@markusschmaus I observe in my branch (test_metamodel_provider_advanced_test3_global_single_metamodel) that for an object of type `Obj` of that grammar
 *  the object data is removed from the obj_attr in `attrs = metamodel.obj_attrs.pop(id(obj_attrs))` (textx/model.py:415), before
 * `metamodel.setattr(obj, attr.name, resolved)` (textx/model.py:851) is called.

Investigating...

goto40:
As far as I understood, the concrete instance creation should happen after all references are resolved... I check the case in case of multi-file (or later multi meta model multi file) cases.

I see no problem to adapt all standard scope providers to adapt to the new object access methods...

goto40:
Hm...
I see that we have the following stages (compare http://textx.github.io/textX/stable/scoping/#using-the-scope-provider-to-modify-a-model), implemented in model.py
 1. **initial raw model w/o references** `model = process_node(parse_tree)`
     * Here, we create user objects `attrs = metamodel.obj_attrs.pop(id(obj_attrs))` and `inst.__init__(**attrs)` (**code A**) - not 100% sure that I got this correctly....
     * Note: object processors for match rules (e.g. regular expressions or rules like 'ID') are called somewhere here, bottom up during model construction. 
 2. **prepare model for full creation** (create a reference resolver) `_start_model_construction(model)`
 3. **load other model files** `scope_provider.load_models(model, encoding=encoding)` (repeat step 1-3, w/o steps 4-6)
 4. **resolve references if all models have been loaded** `if is_main_model: ...`
     * Here we set references/'pointers' to (existing) objects as textx-references into other object attributes `m._tx_reference_resolver.resolve_one_step()`-->`metamodel.setattr(obj, attr.name, resolved)`
     * Note, that the object to be resolved already exists at this point... (**code B**) 
 5. **terminate model creation** `_end_model_construction(m)` (cleanup resolver) for all loaded models
 6. **call obj processors** `call_obj_processors(m._tx_metamodel, m)` for all loaded models

I see a conflict in **code A** and **code B**, if the object is frozen (read-only):
  * The object is created in **code A** and will then be immutable
  * resolving the reference in **code B** violates immutability.

If we move the object creation to a later point (e.g. into step 5):
  * We could solve the problem,
  * However, we have to manage that we create references to proxies (**objects of resolved references do not yet exist at the point of scope resolution...**)

The minor problem of adapting the standard scope providers is less critical (we could make `textx_isinstance` capable of handling the proxies correctly). The main issue for me is the management of proxies and *references* to proxies (which have to be *corrected* when replacing the proxies with the concrete class). 

@markusschmaus, @igordejanovic 
What do you think?
 * The proxies could be a nice enhancement... But we should carefully plan the implementation.
 * Or do we restrict the use of such immutable objects not to contain references?
 * Or any other ideas?
 * Or did I get the idea wrong... (compare the testfile: https://github.com/goto40/textx-playground/blob/master/tests/functional/test_scoping/test_metamodel_provider3_custom_classes.py#L11)

goto40:
I think step 5 may be the correct place to create the instances...?

igordejanovic:
@goto40 I suggest moving the discussion regarding initialization postponing to the #249 to keep it  in the context of the PR.

goto40:
> @goto40 I suggest moving the discussion regarding initialization postponing to the #249 to keep it in the context of the PR.

Sorry: my bad...

goto40:
Very good work @markusschmaus. You may have a look at https://github.com/goto40/textx-playground/commit/f7eafbe707870cb973b506c3e748eb86a20b91c5: 
 * I have a small adaptation in `tools.py`
 * I have an adaptation in `model.py` (@igordejanovic @markusschmaus : I was surprised that the elements were compared field-by-fields... I changed the criteria to `id(obj)`.

You could also incroporate the unittest https://github.com/goto40/textx-playground/blob/master/tests/functional/test_scoping/test_metamodel_provider3_custom_classes.py

I could imagine that we have some minor details missing, but so far it looks good to me!

**affects #249**

igordejanovic:
>  * I have an adaptation in `model.py` (@igordejanovic @markusschmaus : I was surprised that the elements were compared field-by-fields... I changed the criteria to `id(obj)`.

@goto40 Looks good. But I would suggest an optimization for the [`map` transformation](https://github.com/goto40/textx-playground/commit/f7eafbe707870cb973b506c3e748eb86a20b91c5#diff-1d695ebcdd14895ceda26ac647479048R81). It can be a time consuming to call it each time. It is better to introduce another variable `collected_ids` of `set` type and the check would be `if id(elem) in collected_ids`. Each time an element is appended to `collected` its `id` should be added to the `collected_ids` set. For large models this should be a significant speedup.

**affects #249**

igordejanovic:
Also, we should add `attr` package to test dependencies to ensure CI will pick it up also.

**affects #249**

goto40:
OK for explicitly checking for None.

markusschmaus:
I just fixed the pending style error. Can we merge this?

igordejanovic:
@markusschmaus There were still styling errors but since all are taken care of in 174d05e I've merged first two commits of this branch to the #253.

PR #254, opened 2020-05-07
some test and scoping fixes

goto40:
I hope I understood right, that I rebase against this branch...

## Code review checklist

### DIFF
--- a/textx/model.py
+++ b/textx/model.py
@@ -418,24 +418,6 @@ def parse_tree_to_objgraph(parser, parse_tree, file_name=None,
                 _setattr(
                     obj_attrs, 'parent', parser._inst_stack[-1][0])
 
-            # # If the the attributes to the class have been collected in
-            # # _tx_obj_attrs we need to do a proper initialization at
-            # # this point.
-            # if hasattr(obj.__class__, '_tx_obj_attrs'):
-            #     try:
-            #         # Get the attributes which have been collected in
-            #         # metamodel.obj and remove them from this dict.
-            #         attrs = obj.__class__._tx_obj_attrs.pop(
-            #             id(obj))
-            #         inst.__init__(**attrs)
-            #     except TypeError as e:
-            #         # Add class name information in case of
-            #         # wrong constructor parameters
-            #         e.args += ("for class %s" %
-            #                    inst.__class__.__name__,)
-            #         parser.dprint(traceback.print_exc())
-            #         raise e
-
             # Special case for 'name' attrib. It is used for cross-referencing
             if _hasattr(inst, 'name') and _getattr(inst, 'name'):
                 # Objects of each class are in its own namespace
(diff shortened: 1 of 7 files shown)


goto40:
I still have the effect with the missing exception (still investigating).

goto40:
I finally identified the issue with the user class "Model".
 * There were several issues with `_getattr` etc. (including a missing `_delattr`) - I fear it is not the end (but we will reach this point while adding more tests).
 * The most important point was, that the root model needs _tx_filename (and some more aspects maybe when resolving multifile models... - not yet fully analyzed)
 * Moreover `attr` strips leading `_`for parameters when calling `__init__`... For this point we need a good solution to flexibly adapt the parameters... (see **TODO** in  model.py).

@markusschmaus @igordejanovic open for comments on this...

markusschmaus:
I don't see the problem with the skipped underscore, I am using such attributes in some of my classes in my private project and it works fine. I specify the attribute in the grammar without the underscore and the only impact is that this attribute is no longer accessible once the object gets initialized, but this happens only at the end anyway.

However I do think it would be good to allow object processors to change the dictionary passed to `__init__` rather than the initialized object itself. For frozen objects it would be too late otherwise, and this could also help with the in some use cases regarding the underscore.

I was thinking that we might want to add a flag to `pre_init` to `register_obj_processors`, which defaults to `False` and then

textx/model.py
```diff
@@ -658,8 +658,13 @@ def call_obj_processors(metamodel, model_obj,
 
                 for m in models:
                     # TODO: what does this check?
-                    assert not m._tx_reference_resolver.parser._inst_stack
+                    assert not _getattr(m, "_tx_reference_resolver").parser._inst_stack
```
igordejanovic: @goto40 I saw this TODO comment in the code. This line make sure that we have processed all parsed instances kept on `_inst_stack` stack. If at this point there is any instance on the stack we did something wrong.

igordejanovic:
@goto40 
> I hope I understood right, that I rebase against this branch...

Yes, you did. BTW, you can drop commit range 
```
ac50c3d | * Style change for ci test
bc85614 | * quick fixed some individual problems
836dc09 | * Use additional set of ids to avoid relying on __eq__ of user class
ae237ef | * Reverse order of object initialization, so children get initialized before their parent
6c7f394 | * Move initialization before object processors
ccbeacd | * Store attributes of user classes outside object and postpone initialization
```

from this branch as those changes are already included in the `feature/postponed-userobj-init`

igordejanovic:
> I finally identified the issue with the user class "Model".

Great work!

> There were several issues with _getattr etc. (including a missing _delattr) - I fear it is not the end (but we will reach this point while adding more tests).

Yeah, I don't feel good about this API change. We should try to find another, backward compatible solution if possible.

> The most important point was, that the root model needs _tx_filename (and some more aspects maybe when resolving multifile models... - not yet fully analyzed)

Yes, the root of the model is special in that regard that there we keep various meta-data. I would go for a simple solution, prohibiting for a root of the model to be immutable/__slots__ class. Trying to fix that is a lot of work and introducing a lot complexity in the code for supporting a use-case that in my view will be very rare.

> Moreover attr strips leading _for parameters when calling __init__... For this point we need a good solution to flexibly adapt the parameters... (see TODO in model.py).

I don't think that should be our issue if that is the way `attr` works. The user of `attr` should be aware and adapt accordingly. Also, like in the case above, we should always strive for simplicity and never try to find a solution for some rare edge cases in advance.

PR #255, opened 2020-05-07
Msch pre init proxy

markusschmaus:
## Code review checklist

I created a proxy based API for the `_(get|set|has)attr` methods. You can wrap an object which will give you a proxy object which you can read from and write to using the normal syntax `obj.attr`, which gets translated behind the scenes into calls to `_(get|set|has)attr`. Any objects that get returned will equally be wrapped, so `first.second.third` works fine. If the attribute is a list, tuple, or dictionary a special class is used for wrapping in order to implement the full interfaces for these objects. Things that don't work for these wrapped objects are `type(obj)` and `obj.__class__` (use `get_attrs_class(obj)` instead), `obj.__dict__` (use `get_attrs_dict(obj)` instead), as well as `isinstance(obj, something)` (use `textx_isinstance(obj, something)` instead, which has expanded functionality).

I noticed that `textx_isinstance` was often imported locally, and I am unsure if I handled this correct in the instances where I added it. I also didn't check if this was necessary in every instance where I replaced `isinstance` with `textx_isinstance`.

@goto40 @igordejanovic, I would love to hear your feedback

### DIFF
--- a/tests/functional/test_processors.py
+++ b/tests/functional/test_processors.py
@@ -92,6 +92,54 @@ def test_object_processors():
     assert call_order == [2, 2, 2, 1]
 
 
+def test_object_processors_user_classes():
+    """
+    Test that object processors are called.
+    They should be called after each model object construction.
+    """
+
+    def first_obj_processor(first):
+        first._first_called = True
+        first._a_copy = first.a
+
+    def second_obj_processor(second):
+        second._second_called = True
+        second._sec_copy = second.sec
+
+        # test that parent is fully initialised.
+        # b should be True
+        assert second.parent.b is not None
+
+    obj_processors = {
+        'First': first_obj_processor,
+        'Second': second_obj_processor,
+        }
+
+    class First(object):
+        def __init__(self, seconds, a, b, c):
+            self.seconds = seconds
+            self.a = a
+            self.b = b
+            self.c = c
+
+    class Second(object):
+        def __init__(self, sec, parent):
+            self.sec = sec
+            self.parent = parent
+
+    metamodel = metamodel_from_str(grammar, classes=[First, Second])
+    metamodel.register_obj_processors(obj_processors)
+
+    model_str = 'first 34 45 7 A 45 65 B true C "dfdf"'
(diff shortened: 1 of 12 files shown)


igordejanovic:
@markusschmaus Thanks Markus.

Didn't do a deep dive in your implementation so please take these comments with a grain of salt.

If I understand you are wrapping user object into proxies for making a backward compatible API. You are probably not doing unwrapping at the end of the process as that would require reconnecting all references in the object graph.

While proxy wrappers might seems like a good solution they introduce many problems, I think more than they are trying to solve:

1. The biggest issue with proxy wrappers is runtime overhead, both in memory consumption and in speed (due to indirection) which defies the very purpose of using `__slots__` in the first place.

2. Another issue, maybe even more critical is that users end up with objects which are not actually instances of classes they provided.

3. Arguably: complexity

If I misunderstood the implementation please correct me.

markusschmaus:
Objects get wrapped before being handed to to scope providers and unwrapped afterwards.

1. I tried to time the different access methods and it appears that while the `_getattr` function is significantly slower than direct access, the access via the proxy doesn't incur additional slowdown. https://gist.github.com/84b77036eccf90c6f0b2c8e2c43ded56. The wrapper class certainly incurs some memory overhead, but only the objects accessed by the scope provider get wrapped and after it returns garbage collection should take care of it. One simple optimization is of course to change the wrapper classes to `__slots__` classes. Both memory and speed impact only apply while executing the scope providers
2. Objects are only wrapped when passed to scope providers and unwrapped afterwards, so users will indeed get objects of the classes they provided
3. That's a fair argument

igordejanovic:
@markusschmaus Thanks for clarification and detailed analysis.

I'm sorry, I misunderstood the implementation. The proxies are used for scoping providers to provide an easier navigation over the object graph while the object returned to the users are the real objects. Sounds good, and the implementation is great. 

If we don't hit problems I'm currently more in the favor of #256  as the solution is simpler and should be fully backward compatible.

markusschmaus:
Superseded by https://github.com/textX/textX/pull/256

PR #253, opened 2020-05-07
Store attributes of user classes outside object and postpone init

igordejanovic:
Related to #249 #250 #252 #254 #255 #256 #257 #258 

closes #247 

This is a feature branch used to collect various PRs regarding postponing user object initialization needed to support `__slots__` and immutable objects.

/cc @goto40 @markusschmaus

## Code review checklist

- [ ] Pull request represents a single change (i.e. not fixing disparate/unrelated things in a single PR)
- [ ] Title summarizes what is changing
- [ ] Commit messages are meaningful (see [this][commit messages] for details)
- [ ] Tests have been included and/or updated
- [ ] Docstrings have been included and/or updated, as appropriate
- [ ] Standalone docs have been updated accordingly
- [ ] Changelog(s) has/have been updated, as needed (see `CHANGELOG.md`, no need
      to update for typo fixes and such).

[commit messages]: https://chris.beams.io/posts/git-commit/

PR #256, opened 2020-05-08
Postponed userobj init with instrumented user classes

igordejanovic:
I've been working today on a different approach to user obj. initialization postpone.

In order to support postponed initialization, user classes `*attr` methods are
patched during model loading and restored at the end just before `__init__`
methods are called.

Currently this isn't working if same user class is used in multiple meta-models
which depend on each other as the instrumentation is removed while still needed
for referring meta-model (please see test_issue140.py). Looks like the instrumentation should be extended to be local per model. @goto40 please take a look and see if this direction looks good to you.

This PR is based on ongoing #253, first commit is reverting calling `*attr` methods already merged to the feature branch and the second commit is the actual proposed change.

**EDIT**: I have also collected user class instances along the way, so no need for `get_children` call.

/cc @goto40 @markusschmaus

## Code review checklist

### DIFF
--- a/textx/model.py
+++ b/textx/model.py
@@ -305,7 +305,6 @@ def get_model_parser(top_rule, comments_model, **kwargs):
                     pass
             return super(type(obj), obj).__getattribute__(name)
 
-
         def _getattr(obj, name):
             try:
                 return obj._tx_obj_attrs[id(obj)][name]
@@ -340,7 +339,8 @@ def get_model_parser(top_rule, comments_model, **kwargs):
             """
             for user_class in self.metamodel.user_classes.values():
                 if not hasattr(user_class, '_tx_instrumented'):
-                    for a_name in ('getattr', 'setattr', 'delattr', 'getattribute'):
+                    for a_name in ('getattr', 'setattr', 'delattr',
+                                   'getattribute'):
                         real_name = '__{}__'.format(a_name)
                         cached_name = '_tx_real_{}'.format(a_name)
                         setattr(user_class, cached_name,
@@ -361,7 +361,8 @@ def get_model_parser(top_rule, comments_model, **kwargs):
                     user_class._tx_instrumented -= 1
                     if user_class._tx_instrumented == 0:
                         delattr(user_class, '_tx_instrumented')
-                        for a_name in ('getattr', 'setattr', 'delattr', 'getattribute'):
+                        for a_name in ('getattr', 'setattr', 'delattr',
+                                       'getattribute'):
                             cached_name = '_tx_real_{}'.format(a_name)
                             real_name = '__{}__'.format(a_name)
                             if hasattr(user_class, cached_name):
@@ -440,7 +441,7 @@ def parse_tree_to_objgraph(parser, parse_tree, file_name=None,
                         return process_node(
                             next(n for n in node
                                  if type(n) is not Terminal
-                                 and n.rule._tx_class is not RULE_MATCH))
+                                 and n.rule._tx_class is not RULE_MATCH))  # noqa
                     except StopIteration:
                         # All nodes are match rules, do concatenation
                         return ''.join(text(n) for n in node)
--- a/textx/model.py
+++ b/textx/model.py
@@ -12,7 +12,6 @@ from textx.const import MULT_OPTIONAL, MULT_ONE, MULT_ONEORMORE, \
     MULT_ZEROORMORE, RULE_ABSTRACT, RULE_MATCH, MULT_ASSIGN_ERROR, \
     UNKNOWN_OBJ_ERROR
 from textx.lang import PRIMITIVE_PYTHON_TYPES
-from textx.metamodel import _setattr, _getattr, _hasattr
 from textx.scoping import Postponed, remove_models_from_repositories, \
     get_included_models
 from textx.scoping.providers import PlainName as DefaultScopeProvider
@@ -31,8 +30,8 @@ def get_model(obj):
     Finds model root element for the given object.
     """
     p = obj
-    while _hasattr(p, 'parent'):
-        p = _getattr(p, 'parent')
+    while hasattr(p, 'parent'):
+        p = p.parent
     return p
 
 
@@ -57,8 +56,8 @@ def get_parent_of_type(typ, obj):
     if type(typ) is not text:
         typ = typ.__name__
 
-    while _hasattr(obj, 'parent'):
-        obj = _getattr(obj, 'parent')
+    while hasattr(obj, 'parent'):
+        obj = getattr(obj, 'parent')
         if obj.__class__.__name__ == typ:
             return obj
 
@@ -98,11 +97,11 @@ def get_children(decider, root, children_first=False):
                 # Follow only attributes with containment semantics
                 if attr.cont:
                     if attr.mult in (MULT_ONE, MULT_OPTIONAL):
-                        new_elem = _getattr(elem, attr_name)
+                        new_elem = getattr(elem, attr_name)
                         if new_elem:
                             follow(new_elem)
                     else:
-                        new_elem_list = _getattr(elem, attr_name)
+                        new_elem_list = getattr(elem, attr_name)
                         if new_elem_list:
                             for new_elem in new_elem_list:
                                 follow(new_elem)
@@ -415,16 +414,14 @@ def parse_tree_to_objgraph(parser, parse_tree, file_name=None,
 
             # If this object is nested add 'parent' reference
             if parser._inst_stack:
-                _setattr(
-                    obj_attrs, 'parent', parser._inst_stack[-1][0])
+                setattr(obj_attrs, 'parent', parser._inst_stack[-1][0])
 
             # Special case for 'name' attrib. It is used for cross-referencing
-            if _hasattr(inst, 'name') and _getattr(inst, 'name'):
+            if hasattr(inst, 'name') and inst.name:
                 # Objects of each class are in its own namespace
                 if not id(inst.__class__) in parser._instances:
                     parser._instances[id(inst.__class__)] = {}
-                parser._instances[id(inst.__class__)][_getattr(inst, 'name')]\
-                    = inst
+                parser._instances[id(inst.__class__)][inst.name] = inst
 
             if parser.debug:
                 parser.dprint("LEAVING INSTANCE {}".format(node.rule_name))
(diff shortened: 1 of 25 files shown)


markusschmaus:
I like this approach.

goto40:
I need a few days to catch up with the great ideas you (@markusschmaus @igordejanovic) came up. I greatly appreciate the positive dynamics in this project...

Without having analysed #256 at all and without deeping into #255, I found proxies a clear and transparent thing... Looking forward to analyze both approaches!

textx/model.py
```diff
@@ -289,6 +297,62 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+                obj_attrs = obj.__class__._tx_obj_attrs
+                return name in obj_attrs[id(obj)]
+            except AttributeError:
+                return False
+
+        def _getattr(obj, name):
+            try:
+                return obj._tx_obj_attrs[id(obj)][name]
+            except KeyError:
+                raise AttributeError
```
markusschmaus: Some objects of this class may be fully initialized ordinary objects (for example builtins), so we should fall back to the original method if the id is not in this set.

textx/model.py
```diff
@@ -289,6 +297,62 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+                if not hasattr(user_class, '_tx_instrumented'):
+                    for a_name in ('has', 'get', 'set', 'del'):
+                        real_name = '__{}attr__'.format(a_name)
+                        cached_name = '_tx_old_{}attr'.format(a_name)
+                        setattr(user_class, cached_name,
+                                getattr(user_class, real_name, None))
+                        setattr(user_class, real_name,
+                                getattr(self.__class__,
+                                        '_{}attr'.format(a_name)))
+                    setattr(user_class, '_tx_instrumented', True)
```
markusschmaus: We should make this a count which we increase if this method is called for the same class again:
```python
if not hasattr(user_class, '_tx_instrumented'):
    ...
else:
    user_class._tx_instrumented += 1

File: textx/model.py [hunk 3]
```diff
@@ -289,6 +297,62 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+
+        def _restore_user_attr_methods(self):
+            """
+            Restore original has/get/set/del(attr) methods on user
+            classes.
+            """
+            for user_class in self.metamodel.user_classes.values():
+                if hasattr(user_class, '_tx_instrumented'):
```
markusschmaus, line 360: Here we should decrease the count:
````python
if hasattr(user_class, '_tx_instrumented'):
    if user_class._tx_instrumented == 1:
        ...
    else:
        user_class._tx_instrumented -= 1
```

igordejanovic:
@markusschmaus Incredible. I see we came up with exactly the same solution with the last issues (see the last commit) :)

markusschmaus:
😁 Not for the first time. When you suggested using `collected_ids` in `get_children`, I already had that exact code in a local branch.

markusschmaus:
I found instances like this, when providers access `__dict__` on classes which have potentially postponed initialization. https://github.com/textX/textX/blob/e7a544ac6701b42e72c365575c7992a280b13880/textx/scoping/providers.py#L178

We should probably monkey patch `__dict__` for user classes as well using similar code to this. https://github.com/textX/textX/blob/d2dc1f212f524f779b3241cc92ee39ae157251c4/textx/metamodel.py#L130-L134

textx/model.py
```diff
@@ -289,6 +301,85 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+                    user_class._tx_instrumented -= 1
+                    if user_class._tx_instrumented == 0:
+                        delattr(user_class, '_tx_instrumented')
+                        for a_name in ('has', 'get', 'set', 'del'):
```
markusschmaus: There is no `__hasattr__` hook in python, `hasattr` simply calls `getattr` and returns `False` if an `AttributeError` is thrown, and `True` otherwise. So there is no need to monkey patch it.

BTW, this also means that when running `if hasattr(obj, name) and getattr(obj, name) is not None` `obj.__getattr__` gets called twice, which could be avoided using a `try` `except` block, thought this might make the code read a little more clunky.

textx/model.py
```diff
@@ -57,8 +56,8 @@ def get_parent_of_type(typ, obj):
     if type(typ) is not text:
         typ = typ.__name__
 
-    while _hasattr(obj, 'parent'):
-        obj = _getattr(obj, 'parent')
+    while hasattr(obj, 'parent'):
+        obj = getattr(obj, 'parent')
```
goto40: Minor point:`obj.parent`

textx/model.py
```diff
@@ -289,6 +301,85 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+                        real_name = '__{}attr__'.format(a_name)
+                        cached_name = '_tx_real_{}attr'.format(a_name)
+                        setattr(user_class, cached_name,
+                                getattr(user_class, real_name, None))
+                        setattr(user_class, real_name,
+                                getattr(self.__class__,
+                                        '_{}attr'.format(a_name)))
+                    user_class._tx_instrumented = 1
+                else:
+                    user_class._tx_instrumented += 1
```
goto40, line 352: What exactly is the meaning of `_tx_instrumented`? Why is this an integral value and not a boolean flag?

goto40 (commented):
The solution looks very elegant. Great work @igordejanovic and @markusschmaus.

Since the solution mimics the behavior we had with mutable (normal) classes, I do not see a high risk of breaking stuff.

Please have look at#257 concerning the fix of model.py. I think that you could also move the whole object finalization into the method `_end_model_construction` (I will add an online comment concerning this).

goto40:
My overall feeling is very good now! With this, your addition (@markusschmaus) gets very nice and simple to understand - and will allow a whole bunch of new use cases for textX! Thank you.

goto40:
Some details: Should we take care of transforming the attributes for custom objects (before/after object processors)? 

I think @markusschmaus already commented on this ("_However I do think it would be good to allow object processors to change the dictionary passed to `__init__` rather than the initialized object itself. For frozen objects it would be too late otherwise, and this could also help with the in some use cases regarding the underscore._" in #254).

See #257 for a test, where I added a "TODO" (https://github.com/textX/textX/pull/257/commits/3be6282805eddf551cadf7f9410bc5574f10bb36)...: With a User class as model I get
   * not hasattr(m, '_tx_filename')
   * hasattr(m, '_tx_metamodel')

This seems not logical to me (both should be there or not there).

Also I would like to control whether I get these extra fields or not - in the latter case with all consequences.

textx/model.py
igordejanovic: Ah, you are right, good catch. Fixed.

textx/model.py
igordejanovic: Fixed.

textx/model.py
igordejanovic: There is a situation where the same user class might  be used in multiple (meta)model (like in test_issue140). Since instrumentation is not idempotent operation we must be sure to apply it only once and count the number of application to revert the operation when the count drops to 0 (i.e. each application has its undo but they are nested). Another approach would be to use a bool flag and instrument only once and do instrumentation removal at the place which comes after all loading is complete (with all imported models) and we are sure that instrumentation is not needed anymore.

igordejanovic:
> Some details: Should we take care of transforming the attributes for custom objects (before/after object processors)?

> I think @markusschmaus already commented on this ("However I do think it would be good to allow object processors to change the dictionary passed to __init__ rather than the initialized object itself. For frozen objects it would be too late otherwise, and this could also help with the in some use cases regarding the underscore." in #254).

@goto40 @markusschmaus I think that object processing is not that interesting for user classes as all the logic can be put in `__init__` of the provided class. So, in my view, object processors are only interesting for mutable textX provided classes. In a way those two overlap in functionality they provide. Object processors are a lightweight object customization where you don't need a full control over the object type.

igordejanovic:
> We should probably monkey patch `__dict__` for user classes as well using similar code to this. https://github.com/textX/textX/blob/d2dc1f212f524f779b3241cc92ee39ae157251c4/textx/metamodel.py#L130-L134

Agreed. I'll add the patch.

igordejanovic:
What is left for this PR is:

- [x] An inconsistency in model root attributes found by @goto40 in [comment above](https://github.com/textX/textX/pull/256#issuecomment-626309207)
- [ ] Make it work for Python 2 also

goto40:
Great! I just adapted the test in [#257](https://github.com/textX/textX/pull/257/files#diff-de9ce6dc3540692c38ceee8157ce7280R110) to demonstrate (and ensure) that "normal" user classes behave as expected.

igordejanovic:
> See #257 for a test, where I added a "TODO" (3be6282)...: With a User class as model I get
> not hasattr(m, '_tx_filename')
> hasattr(m, '_tx_metamodel')
> This seems not logical to me (both should be there or not there).

@goto40 Good catch. The issue was due to missing `_tx_filename` on a class so in immutable case no attribute could be found while `_tx_metamodel` could as it existed on a class. Fixed by adding `_tx_filename` on a class level but for immutable classes user should be aware that all `_tx_*` attributes are actually on a class level (probably a note should be added in the docs).

I've do the remaining fixes and cleanup of this PR. Also added tests done by @goto40 from PR #257 so that PR can be closed.

This feature seems to be fully backward compatible now but there is still an issue with Python 2.x which handle method bindings differently. Thus, instead of making more kludges in the code to support it I suggest that we target this to the next 3.0 version of textX which will drop support for Python 2. Anybody needing Python 2 still can use textX 2.1.

Will make the announcement of dropping Python 2 support in the upcoming version shortly.

markusschmaus:
I think I just fixed the Python 2 issue (https://github.com/textX/textX/pull/258).

textx/model.py
```diff
@@ -272,23 +271,108 @@ def get_model_from_str(self, model_str, file_name=None, debug=None,
...
+                except (AttributeError, TypeError):
+                    return super(type(obj), obj).__delattr__(name)
+
+        def _replace_user_attr_methods(self):
+            """
+            Replace get/set/del(attr) methods on user classes
+            to support postponing of user obj initialization.
+            """
+            for user_class in self.metamodel.user_classes.values():
+                if not hasattr(user_class, '_tx_instrumented'):
```
markusschmaus: If a user class is a sub class of another user class, which has already been instrumented, then this returns true, even though the class itself still uses its original `attr` methods. I fixed this in https://github.com/textX/textX/pull/258 by checking for `if '_tx_instrumented' not in user_class.__dict__:`

goto40:
If python 2 makes too much problems, I am OK to drop support for it...

@markusschmaus : very good that you fixed the python 2 compatibility.

igordejanovic:
> I think I just fixed the Python 2 issue (#258).

Great!

> If python 2 makes too much problems, I am OK to drop support for it...

Yeah, not only that we are spending some time in keeping backward compatibility it's keeping us from moving forward with introducing Python 3 only features. I particularly look forward to type hinting and static type checking.

PR #258, opened 2020-05-15
Fix infinite recursion

markusschmaus:
## Code review checklist

(Inception PR)

This fixes the infinite recursion we are getting when trying to call super suing type(obj). By defining the replacement methods inside a method each replacement is bound to the original class without needing to look it up.

### DIFF
--- a/tests/functional/test_user_class_attrs_frozen.py
+++ b/tests/functional/test_user_class_attrs_frozen.py
@@ -66,8 +66,8 @@ def test_inheritance_attrs():
 
     @attr.s()
     class Super(object):
+        parent = attr.ib()
         a = attr.ib()
-        parent = attr.ib(kw_only=True, default=None)
 
     @attr.s()
     class Sub(Super):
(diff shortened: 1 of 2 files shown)


goto40 (commented):
Awesome! Can you add a small unittest?

markusschmaus:
Most of this PR addresses a problem with python2 which led all tests on my end fail. I added a small test which tests the correct behavior in the case of attrs classes which inherit from each other, which is also fixed in this PR.

markusschmaus:
I fixed this error:
```
E               ValueError: 

Non keyword-only attributes are not allowed after a keyword-only attribute
```
Newer versions of attrs (including the one I use) no longer have this restriction. https://github.com/python-attrs/attrs/pull/559

igordejanovic:
Thanks @markusschmaus, great work! This looks good to me.

With this change we can publish this in the next minor release and drop Python 2 in the next major release.

I noticed a [leftover docstring](https://github.com/textX/textX/blob/bf5dd7bd491d631285483280b7fd0e0be24aef97/tests/functional/test_user_class_attrs_frozen.py#L22) so you can remove it. BTW, do you want to clean this branch yourself or should I squash merge?

markusschmaus:
Squash merge sounds great to me

PR #260, opened 2020-05-30
Don't cache super class methods when replacing attr methods

markusschmaus:
## Code review checklist

I noticed that when user classes inherited from each other, then the attrs methods weren't always correctly restored, which this PR fixes analogous to https://github.com/textX/textX/pull/256#discussion_r425743468

### DIFF
--- a/tests/functional/test_user_class_attrs_frozen.py
+++ b/tests/functional/test_user_class_attrs_frozen.py
@@ -70,6 +70,14 @@ def test_inheritance_attrs():
     class Sub(Super):
         b = attr.ib()
 
+    super_getattribute = Super.__getattribute__
+    sub_getattribute = Sub.__getattribute__
+
     mm = metamodel_from_str(grammar, classes=[Super, Sub])
     model = mm.model_from_str(modelstr)
     assert model.sub.b == 'something'
+
+    # Make sure that the special methods of both classes have been correctly
+    # restored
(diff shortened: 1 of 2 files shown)


goto40:
Wow. I am really impressed! This is really diving deep into the language... I just played around to make it clear to me.
```
class A:
  def __getattr__(s, n):
    pass

class B(A):
  pass

print(B.__getattr__ is A.__getattr__)
```

**Main question:** Would it not be better to first store all original methods. Once this is done, replace all original methods. (Instead of store/replace them one by one). To me this seems clearer, because the order of replacement is definitively unimportant then. As a side effect, we may get around w/o `__dict__`. 

Just a **minor side question** : does the class object always have a `__dict__`? Probably yes...

goto40:
As far as I understand, we restore more methods, than we had in the original configuration. Am I right? This is the issue you are working on...?

markusschmaus:
Yes, you are right. If a class doesn't define a method `getattr` will return the corresponding method of one of its super classes. If we cache that and "restore" it afterwards the class will now have some method that it originally didn't possess. If we cache all methods before replacing them, then at least that method will be the correct method of the super class, but it's still a method that was initially not directly part of the class.

Yes, classes always have a `__dict__` and I doubt this will ever change.

igordejanovic:
@markusschmaus Good catch. Thanks. LGTM.

PR #261, opened 2020-06-01
postponed attr init: docs added; fixed multifile problem

goto40:
I started this PR to **document** internal aspects (attr replacement) which allows (a) the future _maintainer of the textx-code_ to get detailed info and (b) the _scope-provider-programmer_ to know what happens under the hood (**see scoping.md + figure**).

I also depicted with a simple unittest (test_user_classes_attr_functions.py) that the attr methods are 100% restored. 
  * Maybe @markusschmaus  already included this aspects in his tests. If so, please report, I will delete the new test in that case...
  * Hwoever, I have still a small issue for python 2. **Help is welcome here!** I do not see exactly what the problem is.

Then I realized that something might be missing and I modified an old test to make this effect clear: When loading an **included model** (e.g., via importURI) the loaded model is not the "main model". In that case the **attr methods are not restored**. **I fixed this (see model.py)**.

Comments welcome!

## Code review checklist

- [X] Commit messages are meaningful (see [this][commit messages] for details)
- [X] Standalone docs have been updated accordingly


igordejanovic:
Looks good.

Apparently, in Python 2 when you attach a method to an object (class is also considered an object) it is behind the scenes wrapped in an `instancemethod` type. The solution is on the https://github.com/textX/textX/tree/feature/postponed_init_docs-py2fix branch. It was a simple checking, during restore, if the cached method has `im_func` attribute. If it does it is a wrapper instance method whose `im_func` attribute is the real function being attached to the class.

goto40:
@igordejanovic Thank you for the python2 fix... Once we drop python2 we could rework some parts of the code. OK for me. If you have no further comments you can squash&merge - or I can do it if you prefer.

igordejanovic:
Go ahead and squash merge. Just give it some nice log message that would best describe what is done.

PR #273, opened 2020-08-10
Analysiy and Fix/issue270

goto40:
This is an analysis branch (and possibly the fix branch) of issue 270.

Found so far: the  problem seems to happen, when the base class in used as custom class...

## Code review checklist

- [squash] Commit messages are meaningful (see [this][commit messages] for details)

### DIFF
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -540,12 +540,19 @@ class TextXVisitor(PTNodeVisitor):
                 rule_name
             )
             if cls is not None:
-                assert rule_name not in self.metamodel.user_classes
+                if hasattr(cls, "_tx_obj_attrs"):
+                    delattr(cls, "_tx_obj_attrs")
                 self.metamodel.user_classes[rule_name] = cls
         else:
             cls = self.metamodel.user_classes.get(rule_name)
 
         if cls is not None:
+            if rule_name in self.metamodel._used_rule_names_for_user_classes:
+                raise TextXSemanticError("redefined imported rule"
+                                         + " {}".format(rule_name)
+                                         + " cannot be replaced by a user class")
+            self.metamodel._used_rule_names_for_user_classes.add(rule_name)
+
             # Initialize special attributes
             self.metamodel._init_class(cls, None, node.position,
                                        external_attributes=True)
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -944,7 +944,7 @@ def language_from_str(language_def, metamodel):
                                    TextXVisitor(parser, metamodel))
 
     # Meta-model is constructed. Validate its semantics.
-    metamodel.validate(is_main=(metamodel is None))
+    metamodel.validate()
 
     # Here we connect meta-model and language parser for convenience.
     lang_parser.metamodel = metamodel
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -535,6 +535,12 @@ class TextXVisitor(PTNodeVisitor):
             self.dprint("Creating class: {}".format(rule_name))
 
         # If a class is given by the user use it. Else, create new class.
+        if rule_name not in self.metamodel.user_classes:
+            user_class = self.metamodel.user_classes_provider(
+                rule_name
+            )
+            if user_class is not None:
+                self.metamodel.user_classes[rule_name] = user_class
         if rule_name in self.metamodel.user_classes:
 
             cls = self.metamodel.user_classes[rule_name]
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -540,8 +540,6 @@ class TextXVisitor(PTNodeVisitor):
                 rule_name
             )
             if cls is not None:
-                if hasattr(cls, "_tx_obj_attrs"):
-                    delattr(cls, "_tx_obj_attrs")
                 self.metamodel.user_classes[rule_name] = cls
         else:
             cls = self.metamodel.user_classes.get(rule_name)
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -944,7 +944,7 @@ def language_from_str(language_def, metamodel):
                                    TextXVisitor(parser, metamodel))
 
     # Meta-model is constructed. Validate its semantics.
-    metamodel.validate()
+    metamodel.validate(is_main=(metamodel is None))
 
     # Here we connect meta-model and language parser for convenience.
     lang_parser.metamodel = metamodel
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -535,16 +535,17 @@ class TextXVisitor(PTNodeVisitor):
             self.dprint("Creating class: {}".format(rule_name))
 
         # If a class is given by the user use it. Else, create new class.
-        if rule_name not in self.metamodel.user_classes:
-            user_class = self.metamodel.user_classes_provider(
+        if self.metamodel.user_classes_provider is not None:
+            cls = self.metamodel.user_classes_provider(
                 rule_name
             )
-            if user_class is not None:
-                self.metamodel.user_classes[rule_name] = user_class
-        if rule_name in self.metamodel.user_classes:
+            if cls is not None:
+                assert rule_name not in self.metamodel.user_classes
+                self.metamodel.user_classes[rule_name] = cls
+        else:
+            cls = self.metamodel.user_classes.get(rule_name)
 
-            cls = self.metamodel.user_classes[rule_name]
-
+        if cls is not None:
             # Initialize special attributes
             self.metamodel._init_class(cls, None, node.position,
                                        external_attributes=True)
--- a/textx/metamodel.py
+++ b/textx/metamodel.py
@@ -37,12 +37,14 @@ def metamodel_from_str(lang_desc, metamodel=None, **kwargs):
 
     """
 
+    is_main_metamodel = metamodel is None
+
     if not metamodel:
         metamodel = TextXMetaModel(**kwargs)
 
     language_from_str(lang_desc, metamodel)
 
-    if "metamodel" not in kwargs:
+    if is_main_metamodel:
         metamodel.validate_user_classes()
 
     return metamodel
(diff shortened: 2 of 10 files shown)


goto40:
Also adding the base class to the grammar helps. So **the problem seems to be restricted to used defined base classes not included in the grammar** ...

goto40:
Even if there is not real use case (maybe?) to this, we should understand why this happens...

**correction:** auto-generated user classes are a use case

goto40:
Goal: detect unused classes and raise appropriate error.

@igordejanovic : what do you think is the best way to detect unused user defined classes? Should we add boolean used-flags or can we do it in a more elegant way...?

Note: We should also take care, that the class could be used in an imported grammar.

igordejanovic:
> @igordejanovic : what do you think is the best way to detect unused user defined classes? Should we add boolean used-flags or can we do it in a more elegant way...?
> 
> Note: We should also take care, that the class could be used in an imported grammar

@goto40 If possible I would go with catching the existing exception and throwing the right one. That would go along nicely with "ask forgiveness, not permission" style :)

goto40:
I changed the code to achieve the desired exception. I added comments to link some code fragments in `model.py` to explain how the check for unused user classes works.

Discussion: 

  * we could also instrument all user user classes instead of just instrumenting the user classes used in the grammar (not tried, but I am confident this should work as well). This would allow the use case sketched by @fgro93 in issue 270.
  * But I also fully understand (and favor) the solution sketched by @igordejanovic which I tried to implement in this PR.

@fgro93: Do you think you can easily fix the issue on your side?
  * If not, we should discuss how we can help here... (I think we could add some hook to allow the user case of @fgro93 *and* prevent used user classes in the default case).
  * If yes, I would add a note in the changelog: user classes not used in the grammar are not allowed (BIC).

goto40:
@fgro93: `issue270.py` (changed files) sketches two possibilities to workaround:
  * add a rule corresponding to the unused user class
  * remove the unused user class

goto40:
@igordejanovic I am not 100% sure about unused user classes and imported grammars... I think it should be ok, but this might be worth a second look.

Generally: is the new way of user class initialization tested with grammars importing other grammars...?

Maybe this should be a new issue?

rettenbs:
Just an idea: Instead of a list of classes, the user can also provide a callback. The callback is called for each rule in the grammar. The callback can either return a user defined class or `None` if the default class should be generated.

fgro93:
@goto40 I'm not a fan of adding rules that are not actually part of the grammar. Manually deleting the classes without rules would be the better option here.  We can eliminate some classes automatically by checking if they are `@abstract` classes, sadly the ecore model we have to use does not use  `@abstract` consistently. Thus, still some extra work and maintaining to do.

I like the solution @rettenbs proposed. This would not require any maintaining. A map with all classes could be easily generated.

goto40:
Thank you @rettenbs for your valuable hint. Digging deeper in the code revealed a (possibly) simple solution: By adding an extra flag we can simply request to ignore those unused classes in the model building process (check the unittest). This should fully enable @fgro93's use case.

Maybe let's wait for some more comments from @igordejanovic ... We do not want to add to many possibilities on how to parse models. But I think this case is not too complicated. --> The default behavior is as @igordejanovic proposed; adding the extra flag on model constructions allows to ignore extra (unused) user classes.

goto40:
Note: I have not correctly documented where `_tx_obj_attrs` is set in the present case. I need to rework the comments and check the code again...

igordejanovic:
Hi, sorry for having to wait on my response for so long.

I'm not a big fan of multiple related params (like a list and a flag on how to use the list). I like the idea of @rettenbs to have `classes` param as either a list of classes (what we have now, with adding a proper exception if the class doesn't exist in the grammar) or a callable which gets called to provide the class. This solves the issue reported here as the callable will not get called for unexisting rules. Furthermore, it will provide a greater flexibility as the classes could be created dynamically from the user side. So, I vote for this solution.

I think this change would be backward compatible as the change in handling of unexisting user class should be considered a bug fix.

goto40:
OK. It should be not so difficult. Derails, such as the parameters passed to the callable can be adjusted after the first implementation...

goto40:
Let's discuss this solution. If I understood correctly, the unittest `test_issue270` should sketch the intended usage.

Please take also time to check the concrete implementation. Comments are welcome.

Does it make sense to pass some internal parser/language details to the function, to generate possibly more elaborated/adapted classes?

textx/metamodel.py
```diff
@@ -176,9 +178,16 @@ def __init__(self, file_name=None, classes=None, builtins=None,
 
         # Convert classes to dict for easier lookup
         self.user_classes = {}
+
+        def default_user_classes_provider(name):
+            return None
+        self.user_classes_provider = default_user_classes_provider
```
igordejanovic: Here I suggest just setting `self.user_classes_provider` to `None`.

textx/metamodel.py
```diff
@@ -485,7 +494,16 @@ def validate(self):
...
+        from textx.exceptions import TextXSemanticError
+        for user_class in self.user_classes.values():
+            if not hasattr(user_class, "_tx_obj_attrs"):
+                # It is not a user class used in the grammar
+                # Note: see textx.lang.visit_rule_name, where
+                #    metamodel._init_class is called with
+                #    external_attributes=True.
+                raise TextXSemanticError(
+                    "unexpected: {} seems to be unused in the grammar".format(
+                        user_class.__name__))
```
igordejanovic, line 515: Not completely sure about this check, but lets go with it for now. I would just change the wording of the exception to "{} class is not used in the grammar".

textx/lang.py
```diff
@@ -535,6 +535,12 @@ def visit_rule_name(self, node, children):
             self.dprint("Creating class: {}".format(rule_name))
 
         # If a class is given by the user use it. Else, create new class.
+        if rule_name not in self.metamodel.user_classes:
+            user_class = self.metamodel.user_classes_provider(
+                rule_name
+            )
+            if user_class is not None:
+                self.metamodel.user_classes[rule_name] = user_class
```
igordejanovic: This block of code might be reworked to improve perf., something like this (not tested):
```python
         if self.metamodel.user_classes_provider is not None:
            cls = self.metamodel.user_classes_provider(
                rule_name
            )
            if cls is not None:
                self.metamodel.user_classes[rule_name] = cls
         else:
             cls = self.metamodel.user_classes.get(rule_name)

         if cls is not None:
            # Initialize special attributes
            self.metamodel._init_class(cls, None, node.position,
                                       external_attributes=True)
         else:
            # Create class to collect attributes. At this time PEG rule
            # is not known.
            cls = self.metamodel._new_class(rule_name, None, node.position)

```

igordejanovic (commented):
Great work @goto40. I've left some suggestions.

igordejanovic:
> Does it make sense to pass some internal parser/language details to the function, to generate possibly more elaborated/adapted classes?

Seems OK just to pass rule name. Do you have some specific info in mind?

textx/lang.py
goto40: ok. added an assert.

textx/metamodel.py
goto40: ok

textx/metamodel.py
goto40: ok

goto40:
> Seems OK just to pass rule name. Do you have some specific info in mind?

No. I was just thinking that maybe someone wants to know about the class structure. But you get this info more or less via the `__init__` call.

goto40:
@igordejanovic I added a new test (in `test_user_classes.py`): This shows some problems left to solve:
 * TBC: I reorganized the check for unused user classes (to be not called for metamodels from imported grammars)
 * TODO: It seems we have a problems for user classes for rules redefined in a grammar importing another grammar; The attributes we prepare cannot match both definitions (of both grammars). See unittest `test_user_classes.py` and `B.tx` importing `A.tx`.

goto40:
I propose to detect rules redefined in an importing grammar and prevent (forbid) to provide user classes for those rules.

If this happens, an error will be raised with an appropriate message.

goto40:
* New detection of user classes for redefined rules implemented (via new attribute `metamodel._used_rule_names_for_user_classes` in `lang.py`). This also allows a more explicit check for unused classes (see inline comments).
 * ~~I had to cleanup `delattr(cls, "_tx_obj_attrs")` in `lang.py`and `metamodel.py` for user classes (requried for resued user classes).~~ (code removed)

textx/metamodel.py
goto40: Now we have a separate function to validate user classes. This new validation function is only called for the meta model for the main grammar (not for imported grammars).

goto40: Now, the check has been clarified: we explicitly check for used rule names... This also makes the check more robust, because after a class has been used as user class it gets an extra attribute `_tx_obj_attrs` permanently.

goto40:
Now, I am quite happy with it. @igordejanovic, ready for a second round. No hurry, take your time...

igordejanovic:
@goto40 Looks good. Great work! 

I do have one more suggestion and one concern.

I would suggest moving test_issue270 module to a standard set of tests as we have added a new feature (callable for classes). Maybe to a `test_user_classes_callable`.

The concern I'm having is preventing user class definition for imported and redefined rules. I understand it's not easy to solve and might introduce another problems and performance issues so I guess we can live with it for now.

Fell free to squash merge this after moving the tests.

goto40:
Great. I will do it this week...

goto40:
> The concern I'm having is preventing user class definition for imported and redefined rules. I understand it's not easy to solve and might introduce another problems and performance issues so I guess we can live with it for now.

Yes. This is hard to solve in general.

However, we could allow user defined classes for classes of a specific grammar (including the namespace; so e.g. introduce a user class for B.Thing and for A.Thing in our example. The namespace/grammar name could be passed to the callable... e.g. `f(name, full_name_including_namespace)`.

goto40:
I renamed the test. I will merge the next days if you habe no more comments (@igordejanovic). The issue with the redefined rules can always be solved by not redefining rules (I see no use case where this should not be trivially possible, just by renaming the new rule - am I right?)

PR #274, opened 2020-08-19
RREL implementation

goto40:
As noted in #111 this is a proposal for a possible implementation of the specification of TEP-001.

Minor things have been modified (see rrel.md):
  * limitation of repepated * was removed
  * added ability to mark a RREL to be applied in multiple models (`+m:`, see rrel.md and unittests).

We can discuss things here.
We could also merge this implementation or the tests with another implementation.

Possible start of a review:
 * ComponentsRrel.tx (single file)
 * Data.tx, Flow.tx (multi file, multi model)
 * tests_rrel.py (basic functionality; model navigation with RREL)
 * test_full_qualified_name_rrel_manual.py (strings as scope provider)

## Code review checklist

- [squash] Commit messages are meaningful (see [this][commit messages] for details)
- [x] update textx grammar  to support RREL (`TODO` in code)  -> new issue


igordejanovic:
Great work @goto40 ! At quick overview looks really good.  I'm a bit busy at the moment but will try to do a full review ASAP.

goto40:
No hurry. Take your time.

goto40:
I fixed a minor (but critical) bug: I created a list of sets, but I had only references to the same set in this list...

textx/metamodel.py
```diff
@@ -243,6 +245,7 @@ def __init__(self, file_name=None, classes=None, builtins=None,
         if file_name:
             file_name = os.path.abspath(file_name)
 
+        # Root path will be dir name of the file if loaded from file.
```
igordejanovic: This line looks like accidentally copied.

textx/metamodel.py
```diff
@@ -254,6 +257,10 @@ def __init__(self, file_name=None, classes=None, builtins=None,
 
     def register_scope_providers(self, sp):
         self.scope_providers = sp
+        for k in self.scope_providers.keys():
+            if isinstance(self.scope_providers[k], string_types):
+                self.scope_providers[k] = create_rrel_scope_provider(
+                    self.scope_providers[k])
```
igordejanovic: This could be:
```python
        for k, v in self.scope_providers.items():
            if isinstance(v, string_types):
                self.scope_providers[k] = create_rrel_scope_provider(v)

```

textx/model.py
```diff
@@ -144,10 +144,11 @@ class ObjCrossRef(object):
         position(int): A position in the input string of this cross-ref.
     """
 
-    def __init__(self, obj_name, cls, position):
+    def __init__(self, obj_name, cls, position, local_scope_provider):
```
igordejanovic: `local_scope_provider` needs to be added to the docstring above.

textx/scoping/rrel.py
```diff
@@ -0,0 +1,475 @@
...
+    """
+    from textx import textx_isinstance
+    from textx.scoping import Postponed
+    if isinstance(rrel_tree, string_types):
+        rrel_tree = parse(rrel_tree)
+    assert isinstance(rrel_tree, RRELExpression)
+    rrel_tree = rrel_tree.seq
+    assert isinstance(rrel_tree, RRELSequence)
+    if isinstance(lookup_list, string_types):
+        lookup_list = lookup_list.split(".")
```
igordejanovic: We should probably find another way to split the lookup name as this assume that `.` will always be used which is generally not true.

A quick thought. A grammar rule might provide this info. Something like:
```
FQN[split='.']: ID+['.'];
```

igordejanovic (commented):
@goto40 I did an incomplete review pass today. Excellent work! Will need more time as this is a big change. See some comments bellow.

textx/metamodel.py
goto40: line removed

textx/metamodel.py
goto40: code changed

textx/model.py
goto40: docstring corrected.

textx/scoping/rrel.py
goto40: Very good idea!

goto40: A first experiment (to be discussed):
  * Allow the `split` parameter
  * ~~Change `create_rrel_scope_provider` to accept a split_string directly (e.g. '.' or '::') or a rule_name used to extract the split_string from the meta model, once all rules are parsed.~~
  * Change `create_rrel_scope_provider` to accept a split_string directly (e.g. '.' or '::') or None. If None is passed the split_string is extracted from the gramme (param "split")
  * Unittests adapted:
     * `test_referencing_attributes_with_rrel_all_in_one_splitstring`: grammar/rule based spli string extraction
     * `test_rrel.py`: find-API test with fixed split string.

Open: 
  * ~~The grammar/RREL expression could contain an option to specify the split_string, like `+m:` for importURI, `+split('::'):`.~~

goto40:
A first version of the split-param is implemented.

textx/scoping/rrel.py
igordejanovic: @goto40 Excellent! I have another proposal: since we have now `+m` as a expression marker which changes the lookup algorithm, we should probably to the same with `^` as they belong to the same kind of operators. Maybe instead of `^` to use something like `+u` for upward search so we could say: `+m+u:...` instead of `+m:^...`. I think this will be more consistent and will open a possibility to introduce more expression markers in the future. So all expression markers will be of the same form `+<letter>` and will come before `:`. 

One more thing to discuss is `~` and `*`. One is prefix and the other is in postfix form. Should we leave it that way? Would it be more consistent to use postfix form for both (or prefix)? Instead of `~extends*` to write `extends~*` (or `extends*~`)? What is more logical, easier to comprehend?

goto40: In general, I agree. However, for '^' we have a different case: you can have it for one path in a sequence and not for another. So it is not really global. Note: '^` is only a shortcut for' (..)*'. 

For `~` : if we use a postfix `(... )*~` is difficult to interpret...

goto40: Of we could later add options like ignore-case, etc.

igordejanovic: Ok, sounds good. Agree on both points.

goto40:
BTW. The implementation of the `rrel.find` function was my first experience, where I found a use case of backtracking and coroutines, __in combination__. That allowed parallel call stacks to coexist, which enables the non-greedyness of the `*` operation. 

Maybe this is standard in some way, for me it was a novel experience! It really simplified things. I am looking for some literature or blogs about that...

textx/lang.py
```diff
@@ -612,7 +623,7 @@ def visit_rule_ref(self, node, children):
         rule_name = text(node)
         # Here a name of the meta-class (rule) is expected but to support
         # forward referencing we are postponing resolving to second_pass.
-        return RuleCrossRef(rule_name, rule_name, node.position)
+        return RuleCrossRef(rule_name, rule_name, node.position, None)  # TODO. check none! what is the difference to the other constructor call below?
```
igordejanovic: `None` seems ok here.

textx/lang.py
```diff
@@ -762,6 +773,8 @@ def visit_assignment(self, node, children):
         if type(rhs_rule) is tuple and rhs_rule[0] == "obj_ref":
             cls_attr.cont = False
             cls_attr.ref = True
+            cls_attr.local_scope_provider = rhs_rule[1].local_scope_provider
+            cls_attr.local_scope_provider_match_rule_name = rhs_rule[1].rule_name
             # Override rhs by its PEG rule for further processing
             rhs_rule = rhs_rule[1]
```
igordejanovic, line 777: It would be slightly more readable to move this unpacking of `rhs_rule` before using it in lines 776 and 777 above.

textx/model.py
```diff
@@ -553,8 +558,12 @@ def process_node(node):
 
                 if metaattr.ref and not metaattr.cont:
                     # If this is non-containing reference create ObjCrossRef
+                    p = metaattr.local_scope_provider
+                    rn = metaattr.local_scope_provider_match_rule_name
```
igordejanovic: Could we make this attribute name shorter? Maybe just `match_rule_name`. Also, maybe `local_scope_provider` -> `scope_provider` (it is understood from the context that it is local).

textx/model.py
```diff
@@ -142,12 +142,17 @@ class ObjCrossRef(object):
         obj_name(str): A name of the target object.
         cls(TextXClass): The target object class.
         position(int): A position in the input string of this cross-ref.
+        local_scope_provider(scope provider): A scope provider for that
+            reference (see scoping.md for requirements of a scope provider)
     """
 
-    def __init__(self, obj_name, cls, position):
+    def __init__(self, obj_name, cls, position, local_scope_provider,
+                 match_rule_name):
```
igordejanovic, line 153: `match_rule_name` needs docstring.

docs/rrel.md
```diff
@@ -0,0 +1,191 @@
...
+  all elements should be processed. For example, to search for a method in the
+  inheritance hierarchy one would write `~extends*.methods` which (due to `*`,
+  see below) first searches `methods` collection of the current context object,
+  if not found, all elements of the current `extends` collection are iterated in
+  the order of defintion without consuming name part, and then name would be
+  searched in the `methods` collection of each object from the `extends`
+  collection. If not found `*` would expand `extends` to `extends.extends` if
+  possible and the search would continue.
+- `*` - Repeat/expand. Used in expansion step to expand sub-expression by 0+
+  times. First expansion tried will be 0, than once, then twice etc. For
```
igordejanovic: Typo 'than' -> 'then'

tests/functional/examples/test_hierarchical_data_structures_referencing_attributes.py
```diff
@@ -153,3 +154,118 @@ def ref_scope(refItem, myattr, attr_ref):
...
+            references+=Reference;
+        Struct:
+            'struct' name=ID '{' vals+=Val '}';
+        Val:
+            'val' name=ID (':' type=[Struct])?;
+        Instance:
+            'instance' name=ID (':' type=[Struct])?;
+        Reference:
+            'reference' instance=[Instance]
+            '.' ref=[Val|FQN|.~instance.~type.vals.(~type.vals)*];
```
igordejanovic: This could be:
```
        Reference:
            'reference' ref=[Val|FQN|instances.~type.vals.(~type.vals)*];
```
which yields a better errors bellow (`Unknown object "c.b.a.not_there"') as the whole path is reported.
Unless you wanted to demonstrate searching from a relative AST position with global instance scope.

tests/functional/test_scoping/test_rrel.py
```diff
@@ -0,0 +1,284 @@
...
+    #################################
+    # TEST
+    #################################
+
+    P2 = find(my_model, "P2", "packages")
+    assert P2 is not None
+    Part2 = find(my_model, "P2.Part2", "packages.classes")
+    assert Part2 is not None
+    rec = find(my_model, "P2.Part2.rec", "packages.classes.attributes")
+    assert rec is not None
```
igordejanovic: Maybe it would be better to assert the name or the type of the found object in all these asserts (`assert rec.name == 'rec'`). It would be a stronger assert.

docs/rrel.md
```diff
@@ -0,0 +1,191 @@
+Changes:
```
igordejanovic: This file needs entry in mkdocs.yml

igordejanovic (commented):
@goto40 Excellent work! I'm really happy how RREL feature is shaping up. Much easier to use and understand comparing to Python based scoping providers. I'm looking forward to start using this.

Please see some minor comments/changes bellow.

igordejanovic:
@goto40 One more suggestion. It is internal stuff so doesn't need to be done in this first implementation but I think that the `find` method could be made simpler if the AST nodes would do searching by recursive find calls. Each node can implement its search logic and delegate search to its subnodes. Also, all RREL AST nodes could inherit some base class for common behavior. The reason why `find` should be simplified is that it relies too much on type checking by `isinstance` and then choose logic based on AST node type.  It is just what cross my mind so definitely need more thinking.

textx/lang.py
goto40: ok."todo" removed

textx/lang.py
goto40: ok.

goto40: ok.

textx/model.py
goto40: done. It is much more readable now...

textx/model.py
goto40: done.

docs/rrel.md
goto40: done.

tests/functional/examples/test_hierarchical_data_structures_referencing_attributes.py
goto40: Great!

goto40: Test changed. Much nicer!

tests/functional/test_scoping/test_rrel.py
goto40: Changed the test: now, we test the name instead of "not None". Also removed redundant tests for "not None", if we access an attribute in an assert afterwards... Added some more checks to make the tests even stronger.

docs/rrel.md
goto40: added entry.

goto40: Also removed the "changes" section. The removed text is here:
```
Changes:
 * ~~Due to expansion of `*` for now we should prohibit using of `*` more than once.~~
 * Added preliminary syntax to enable multi files search  (including ImportURI activation): Use
   the prefix `+m:` for an RREL expression: Then, in case of no match, other models are searched.
   ([grammar example](https://github.com/textX/textX/blob/master/tests/functional/registration/projects/data_dsl/data_dsl/Data.tx))
 * Added docu concerning the pslit param
 * Removed "See [this comment](https://github.com/igordejanovic/textX/issues/111#issuecomment-441308211)"
```

goto40:
> @goto40 One more suggestion. It is internal stuff so doesn't need to be done in this first implementation but I think that the `find` method could be made simpler if the AST nodes would do searching by recursive find calls. Each node can implement its search logic and delegate search to its subnodes. Also, all RREL AST nodes could inherit some base class for common behavior. The reason why `find` should be simplified is that it relies too much on type checking by `isinstance` and then choose logic based on AST node type. It is just what cross my mind so definitely need more thinking.

I really like the idea. It somehow made everything clearer. However, the concrete implementation I tried (https://github.com/textX/textX/pull/274/commits/68dff417a30241fd0ea62cd9fe8640bf1dfbdd33) seems less transparent to me, since the logic is distributed across the code now... I am not sure if this is a clarification or not. To be discussed. We can always revert the code.

Note, if we revert: the handling of "visited" is better in the new version (it is always checked when entering a new state, not only for the '*'-case).

igordejanovic:
@goto40 Excellent work. I like the new implementation more as there is a clear separation of concerns and logic regarding different RREL operators. Adding new operator would be a mere introducing of new RRELBase inherited class. I think this is ready to be merged, we can work further on this if needed in other PRs (e.g. updating textx.tx). 

So, please go ahead and reorganize/clean commits on this branch how you see fit, merge to master and update CHANGELOG accordingly.

Issue #275, opened 2020-08-26
Syntax error in generated dot file


hlouzada:
The generated visualization dot file contains an syntax error if the character `<` or `>` is placed inside an HTML table within an label.

Steps to reproduce:
Run this exemple modified from https://textx.github.io/textX/stable/visualization/
```py
from textx import metamodel_from_str
from textx.export import metamodel_export


grammar = '''

Comment:
    "<:" ( !"*/" /./ )* ":>"
;

EntityModel:
  entities+=Entity    // each model has one or more entities
;

Entity:
  'entity' name=ID '{'
    attributes+=Attribute     // each entity has one or more attributes
  '}'
;

Attribute:
  name=ID ':' type=[Entity]   // type is a reference to an entity. There are
                              // built-in entities registered on the meta-model
                              // for primitive types (integer, string)
;
'''

entity_mm = metamodel_from_str(grammar)

metamodel_export(entity_mm, 'entity_model.dot')
```
Now execute:
```
$ dot -Tpng -O entity_model.dot
```

When parsed _graphviz_ will return an syntax error and that will be solved only if all `>` and `<` are removed. An possible solution is to replace them with their respective Unicode character as `&#62;` and `&#60;` respectively.

OS: Ubuntu 20.04 LTS
Python: Python 3.8.5
textX: 2.2.0

igordejanovic:
@hlouzada Thanks for the detailed report.

goto40:
@hlouzada : the bug should be fixed. You can have a look at the current master.

hlouzada:
Thanks for the fix. It's working now.

PR #276, opened 2020-08-27
Bugfix for issue275 (fix for 'syntax error generated')

goto40:
Analysis branch for #275.
Proposed fix: 
 * introduced a `escale_html` for the table.
 * refactored export to separate html table export from the rest (to be tested separately)
 * added test including a simple html format check

Note:
 * For Python 2.7: the `escape_html` is only implemented rudimentary
 * Python 3.0 - 3.2 is not supported (ok)

## Code review checklist

- [n/a] Standalone docs have been updated accordingly

### DIFF
--- a/tests/functional/regressions/test_issue275.py
+++ b/tests/functional/regressions/test_issue275.py
@@ -37,13 +37,9 @@ def test_issue275():
     metamodel_export_tofile(entity_mm, out_file, renderer)
 
     text = out_file.getvalue()
-    print(text)
     assert text.__contains__("&lt;")
     assert text.__contains__("&gt;")
 
     table = renderer.get_match_rules_table()
-    print("----------")
-    print(table)
-    print("----------")
     html5parser = html5lib.HTMLParser(strict=True)
     html5parser.parseFragment(table)
--- a/tests/functional/regressions/test_issue275.py
+++ b/tests/functional/regressions/test_issue275.py
@@ -5,21 +5,21 @@ import io
 
 def test_issue275():
     grammar = '''
-    
+
     Comment:
         "<:" ( !"*/" /./ )* ":>"
     ;
-    
+
     EntityModel:
       entities+=Entity    // each model has one or more entities
     ;
-    
+
     Entity:
       'entity' name=ID '{'
         attributes+=Attribute     // each entity has one or more attributes
       '}'
     ;
-    
+
     Attribute:
       name=ID ':' type=[Entity]   // type is a reference to an entity. There are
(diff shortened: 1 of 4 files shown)


goto40:
@igordejanovic ready for review

tests/functional/regressions/test_issue275.py
```diff
@@ -0,0 +1,49 @@
...
+
+    entity_mm = metamodel_from_str(grammar)
+
+    out_file = io.StringIO()
+    # metamodel_export(entity_mm, "test.dot")
+    renderer = DotRenderer()
+    metamodel_export_tofile(entity_mm, out_file, renderer)
+
+    text = out_file.getvalue()
+    print(text)
```
igordejanovic: @goto40 We shouldn't be doing printing from tests

tests/functional/regressions/test_issue275.py
```diff
@@ -0,0 +1,49 @@
...
+    assert text.__contains__("&lt;")
+    assert text.__contains__("&gt;")
+
+    table = renderer.get_match_rules_table()
+    print("----------")
+    print(table)
+    print("----------")
```
igordejanovic: Here also.

textx/export.py
```diff
@@ -103,6 +103,14 @@ def dot_escape(s):
             .replace('?', r'\?')
 
 
+def html_escape(s):
+    if sys.version < '3':
+        return s.replace("<", "&lt;").replace(">", "&gt;")
+    else:
+        from html import escape
+        return escape(s)
+
```
igordejanovic, line 113: This is a minor and won't make much difference in this case but it is usually a good thing to do something like:
```python
if sys.version < '3':
    def html_escape(s):
        return s.replace("<", "&lt;").replace(">", "&gt;")
else:
    def html_escape(s):
        from html import escape
        return escape(s)
```

As the condition will be evaluated only once on module import, not every time the func is called.

igordejanovic (commented):
@goto40 Very good. Only some minor comments.

tests/functional/regressions/test_issue275.py
goto40: done

tests/functional/regressions/test_issue275.py
goto40: done

textx/export.py
goto40: done

igordejanovic (approved):
@goto40 LGTM. Ready for squash merge.

PR #284, opened 2020-10-10
Builtin models

igordejanovic:
This PR has two commit. First commit is simplification of model repo access. Second commit is a new feature that adds `builtin_models` param to meta-model which serves similar purpose as `builtins` but is of `ModelRepository` type and enables adding a regular textX models as a fallback for all `ImportURI` based scoping providers.

I'm working on a language where I have two meta-models. One is intended for end-user while the other is internal meta-model used to specify types/object referenced from the user's models. I wanted to free users of having to import these provided specifications. Actually, those spec would be deployed by the language library so it would be hard for users to know their locations, thus it is nice to be able to pre-load those specs during main meta-model construction.

## Code review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ please take a look at related PRs and issues and see if the change affects you.
 
 ### Added
 
+  - `builtin_models` of type `ModelRepository` to meta-model constructor. Used
+    to supply pre-loaded models for `ImportURI` based scoping providers as a
+    fallback to search into. ([#284])
   - Initial implementation of TEP-001 ([#111]) allowing to specify scope
     provider behavior within the grammar itself. [#274] introduces
     the RREL (reference resolving expression language) to define how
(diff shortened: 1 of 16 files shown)


goto40:
I like the `builtin_models`!

I am not sure if we have some double-functionality here: have a look at https://github.com/textX/textX/blob/master/tests/functional/test_metamodel/test_multi_metamodel_refs.py#L49. In this test we use the `global_repository` parameter of `metamodel_from_*`. This seems to have a similar effect. However, I prefer the way you propose it now. Maybe we can get rid of the `global_repository` and map it internally to the same thing? Or at least deprecate `global_repository`...

goto40:
Ok for the simplification of the model repo access. Very good!

goto40:
I think the new feature is a common fallback (lookup), where as the global repo is really global (shared and modified)...

igordejanovic:
In my understanding `global_repository` has different effect. It enables caching of all loaded models by the meta-model to ensure that all objects are unique no matter where they are loaded, as long as the same meta-model is used.  IIRC global repo style scope providers will store all loaded models to global repository `all_models` and if model is loaded again (even through importURI) the cached version will be used.

On the other hand `builtin_models` is meant to be a pre-loaded models for reference resolution and it affects all `ImportURI` inherited scope providers. These models are prepared during construction of the meta-model and are not updated by scope providers. So, in a way this is `builtins` functionality but with proper textX models loaded from str/file.

PR #297, opened 2020-10-29
WIP: Parallelization unit tests (thread- and process-based)

stanislaw:
## Code review checklist

### DIFF
--- a/tests/functional/test_parallelization.py
+++ b/tests/functional/test_parallelization.py
@@ -1,0 +1,91 @@
+from __future__ import unicode_literals
+
+import concurrent.futures
+import multiprocessing
+
+from textx import metamodel_from_str
+
+
+# Define custom classes because default classes provided by textX are not
+# picklable:
+# _pickle.PicklingError: Can't pickle <textx:Second class at 140337157533528>: attribute lookup TextXClass on textx.metamodel failed
+class Main(object):
+    def __init__(self, seconds):
+        self.seconds = seconds
+
+
+class Second(object):
+    def __init__(self, parent, value):
+        self.parent = parent
+        self.value = value
+
+
+grammar = """
+Main:
+    'first' seconds+=Second
+;
+
+Second:
+    value=INT
+;
+
+"""
+
+
+def parse_document_with_threads(n):
+    meta_model = metamodel_from_str(grammar, classes=[Main, Second])
+
+    document = meta_model.model_from_str('first 1 2 3 4')
+    assert len(document.seconds) == 4
+
+    return document
+
+
+def parse_document_with_processes(n):
+    meta_model = metamodel_from_str(grammar, classes=[Main, Second])
+
+    document = meta_model.model_from_str('first 1 2 2 3')
+    assert len(document.seconds) == 4
+
+    # HACK:
+    # ProcessPoolExecutor doesn't work because of non-picklable parts
+    # of textx. The offending fields are stripped down because they
+    # are not used anyway when writing using our generator.
+    document._tx_attrs = None
+    document._tx_parser = None
+    document._tx_attrs = None
+    document._tx_metamodel = None
+    document._tx_model_params = None
+    document._tx_peg_rule = None
+
+    return document
+
+
+# The test always fails here:
+#             # If the the attributes to the class have been
+#             # collected in _tx_obj_attrs we need to do a proper
+#             # initialization at this point.
+#             for obj in the_parser._user_class_inst:
+#                 try:
+#                     # Get the attributes which have been collected
+#                     # in metamodel.obj and remove them from this dict.
+#                     attrs = obj.__class__._tx_obj_attrs.pop(id(obj))
+#                     KeyError: 4346391576
+def test_parallelization_with_threads():
+    # WIP: might be easier to debug this before the code gets threaded
+    parse_document_with_processes(1)
+
+    cpus = multiprocessing.cpu_count()
+    with concurrent.futures.ThreadPoolExecutor(max_workers=cpus) as executor:
+        for document in executor.map(parse_document_with_threads, range(1, 101)):
+            assert len(document.seconds) == 4
+
+
+def test_parallelization_with_processes():
+    # WIP: might be easier to debug this before the code gets threaded
+    parse_document_with_processes(1)
+
+    cpus = multiprocessing.cpu_count()
+    with concurrent.futures.ProcessPoolExecutor(max_workers=cpus) as executor:
+        for document in executor.map(parse_document_with_processes, range(1, 101)):
+            assert len(document.seconds) == 4
--- a/textx/lang.py
+++ b/textx/lang.py
@@ -920,7 +920,6 @@ class TextXVisitor(RRELVisitor):
 # parser object cache as a thread local storage (i.e. cached per each thread).
 # To speed up parser initialization (e.g. during imports)
 tlocal = threading.local()
-tlocal.textX_parsers = {}
 
 
 def language_from_str(language_def, metamodel, file_name):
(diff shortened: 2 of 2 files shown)


stanislaw:
@igordejanovic your initialization of the tlocal's property `textX_parsers` didn't work correctly which I seem to have fixed. However, there is still some shared state in action if I understand it correctly:

```
            for obj in the_parser._user_class_inst:
                try:
                    # Get the attributes which have been collected
                    # in metamodel.obj and remove them from this dict.
>                   attrs = obj.__class__._tx_obj_attrs.pop(id(obj))
E                   KeyError: 4473541184
```

I have added both unit tests. I hope that this will be useful for you to investigate and test.

goto40:
Very good. Do we have any other shared structures (any attribute of the meta model... )? For C++, I would run a thread checker like helgrind to check for racing conditions... 

I could also imagine problems with global model repositories... We could simply state that we have "limited support for parallel execution of `model_from_*` (excluding global model repositories)"?

goto40:
Thank you for your contribution and support!

igordejanovic:
@stanislaw Ah, good catch. Thanks for the investigation and unit tests. It will be much easier to debug.

igordejanovic:
We have module/class level mutable objects in multiple places which should be converted to thread local storage for multi-threading to work correctly. The aim of this project was never focused  on multi-threading/processing, but given the direction we are going in increasing the number of CPU cores, we should definitely put some effort in that direction. This will require some careful rework and testing. The worst thing about the issue is that tests are non-deterministic, so they can't be automatized nor they can be trusted. Any idea on how to approach the testing?

stanislaw:
> The worst thing about the issue is that tests are non-deterministic, so they can't be automatized nor they can be trusted. Any idea on how to approach the testing?

Can you explain in a little bit more detail? Scheduling 100 threads or processes and doing parsing with textx seems safe enough to be included in a test suite. Which problems do you see there?

In the [Mull project](https://github.com/mull-project/mull) we are using two groups of tests:

- Normal unit tests that also include a subset where a fork-based execution of code in subprocesses is tested
- Integration tests based on LIT/FileCheck (https://llvm.org/docs/CommandGuide/lit.html and https://github.com/mull-project/FileCheck.py).

By its design, Mull has a lot of fork-based executions in subprocesses. However, we seem to have very stable test suites by now.

igordejanovic:
> Can you explain in a little bit more detail? Scheduling 100 threads or processes and doing parsing with textx seems safe enough to be included in a test suite. Which problems do you see there?

AFAIK These kind of tests fail at different places with different exceptions. It might happen from time to time to even pass without issues. With a large number of mutable shared objects where things can go wrong they would fail most of the time (that's what happens now, but we cannot assert how they are failing) but as we start cleaning things up they will start to fail more rarely which makes them unreliable but I guess that's something we'll have to live with. I don't believe there is any other option.

stanislaw:
Related comment: with my limited knowledge of Python, I think that the API I am using in this PR does not result in run-away code:

```
     with concurrent.futures.ThreadPoolExecutor(max_workers=cpus) as executor: 
```

it doesn't matter which executor is used: thread-based or process-based - this API always waits until all threads or subprocesses finish their execution and only then the executing thread continues. With this guarantee, it seems pretty safe to include such tests into a PyTest test suite.

---

Maybe it is useful to provide my use case for the parallelization of textX. I am building a tool for requirements and specifications documents management. These documents are written in a grammar created and parsed with textX and some of these documents may contain up to 1000 and more requirement items (the document size can reach few hundreds kilobytes up to few megabytes for a safe margin). The documents of such a size, when there is a number of them in a documentation tree, slow down the export process considerably and this is why I am trying to parallelize textx as well as implement the incremental generation when only the recently touched documents are re-generated.

stanislaw:
> > Can you explain in a little bit more detail? Scheduling 100 threads or processes and doing parsing with textx seems safe enough to be included in a test suite. Which problems do you see there?
> 
> AFAIK These kind of tests fail at different places with different exceptions. It might happen from time to time to even pass without issues. With a large number of mutable shared objects where things can go wrong they would fail most of the time (that's what happens now, but we cannot assert how they are failing) but as we start cleaning things up they will start to fail more rarely which makes them unreliable but I guess that's something we'll have to live with. I don't believe there is any other option.

On my end, I am sure that I will be testing textX with a large number of sufficiently large documents so there is at least one user who will use textx and its parallelization feature at scale 😄 

It might be practical to exclude these tests from the main test suite for now until there is a good confidence in the overall thread safety (to avoid fighting with flaky CI when tests fail randomly). Otherwise, I would be willing to provide a testing hand for textX on its way towards thread-safety.

igordejanovic:
> It might be practical to exclude these tests from the main test suite for now until there is a good confidence in the overall thread safety (to avoid fighting with flaky CI when tests fail randomly). 

Fully agree.

>Otherwise, I would be willing to provide a testing hand for textX on its way towards thread-safety.

Thanks, since I personally don't have a great need at the moment for parallelization it might take awhile until I find some time to put in this area so any help in code changes and testing is greatly appreciated :)

igordejanovic:
> Maybe it is useful to provide my use case for the parallelization of textX.

@stanislaw BTW, very interesting use case. Seems that your project will be a great test-bed for textX performance as well :smiley:

stanislaw:
Doing a quick scan over the shared variables, I have tried to un-globalize some of them starting from `_tx_obj_attrs`. It is indeed a non-trivial effort and requires doing them one by one.

One thing that I noticed: please compare the methods `_getattribute` and `_setattr(obj, name, value):` - the get method accesses `user_class._tx_obj_attrs[...]` and the set method does `obj._tx_obj_attrs[...]`. Shouldn't the set method also access `user_class.` instead of `obj.`?

Check the following:

```py
             def _getattribute(obj, name):
                 if name == '__dict__':
                     try:
                         return user_class._tx_obj_attrs[id(obj)]
                     except KeyError:
                         pass
                 else:
                     try:
                        return user_class._tx_obj_attrs[id(obj)][name]
                     except KeyError:
                         pass
```

 ```diff
             def _setattr(obj, name, value):
                 try:
-                    obj._tx_obj_attrs[id(obj)][name] = value
+                    user_class._tx_obj_attrs[id(obj)][name] = value
                 except KeyError:
```

**Comment by goto40** [comment]
> I could also imagine problems with global model repositories... 

See https://github.com/textX/textX/blob/master/tests/functional/test_scoping/test_global_import_modules.py#L62

**Comment by igordejanovic** [comment]
> Doing a quick scan over the shared variables, I have tried to un-globalize some of them starting from `_tx_obj_attrs`. It is indeed a non-trivial effort and requires doing them one by one.
> 

Yeah, it is a not-trivial thing to do and I wonder is it worth it. Due to Python GIL we won't get any performance improvement and might even get a degradation. My proposal is to state in the docs that textX is not thread-safe, as @goto40 suggested, and should be used with a proper synchronization and focus instead on a proper multi-processing support which could bring a significant performance gain in multi-core environments.

> One thing that I noticed:...

Good catch, thanks. It is inconsistent. Fixed on the master branch.

**Comment by goto40** [comment]
Note: you can always instantiate manually multiple instances of a meta model and use them in parallel...

**Comment by stanislaw** [comment]
I don't think I have enough time to dig into textX internals myself to help you guys. I am not blocked by this feature anyway so there is no urgency from my side.

Just one last comment on the possible implementation: instead of using changing global or class-level variables to local storage, consider assigning them as properties of the meta model object so that the caching is done per meta-model, not per thread.

Two reasons:

1) This threading.local API does not support at least two things: `values()` and `len()` will not work. So you cannot simply replace global {} lists with threading.local. 
2) Using thread-local storage is still a kind of global state, just sliced by threads. A per-metamodel storage makes the whole code to avoid any kind of global variables including those based on the thread-local storage.

Hope this helps and thanks again for the whole TextX effort.

**Comment by stanislaw** [comment]
Forgot to mention this: while the thread-based parallelization does not work because of global variables, the process-based one `test_parallelization_with_processes` actually works for me on macOS.

Of course, these hacks have to be applied:

```
def parse_document_with_processes(n):
    meta_model = metamodel_from_str(grammar, classes=[Main, Second])

    document = meta_model.model_from_str('first 1 2 2 3')
    assert len(document.seconds) == 4

    # HACK:
    # ProcessPoolExecutor doesn't work because of non-picklable parts
    # of textx. The offending fields are stripped down because they
    # are not used anyway when writing using our generator.
    document._tx_attrs = None
    document._tx_parser = None
    document._tx_attrs = None
    document._tx_metamodel = None
    document._tx_model_params = None
    document._tx_peg_rule = None

    return document
```

PR #306, opened 2020-11-29
Transform non-TextXErrors from object processors to TextXErrors

goto40:
Use case: Exceptions from objects processors shall always be TextXErrors, to indicate filename and error location.
```
 grammar = r"""
        Model: date=Date;
        Date: d=INT '.' m=INT '.' y=INT;
        """

    from datetime import datetime
    mm.register_obj_processors({'Date': lambda x: datetime(day=x.d, month=x.m, year=x.y)})
```

## Code review checklist

- [x] Squash commit messages and separate changes from tests/functional/test_user_classes.py ("enhancement of the existing test") 
- [X] Docstrings have been included and/or updated, as appropriate

### DIFF
--- a/tests/functional/test_processors.py
+++ b/tests/functional/test_processors.py
@@ -4,7 +4,7 @@ Testing model and object processors.
 from __future__ import unicode_literals
 import pytest  # noqa
 import sys
-from textx import metamodel_from_str
+from textx import metamodel_from_str, textxerror_wrap
 
 if sys.version < '3':
     text = unicode  # noqa
@@ -350,3 +350,60 @@ def test_multipart_nested_match_rules():
     assert called[0]
     assert model.objects[0] == '#3.4'
     assert model.objects[1] == '--6--'
+
+
+def test_obj_processor_exception_wrap_for_external_exceptions():
+    grammar = r"""
+        Model: date=Date;
+        Date: d=INT '.' m=INT '.' y=INT;
+        """
+
+    @textxerror_wrap
+    def date_converter(model_date):
+        from datetime import datetime
+        return datetime(day=model_date.d, month=model_date.m, year=model_date.y)
+
+    mm = metamodel_from_str(grammar)
+    mm.register_obj_processors({'Date': date_converter})
+    mm.model_from_str('1.12.20')  # ok
+    from textx import TextXError
+    with pytest.raises(TextXError, match=r'None:1:1:.*1\.\.12'):
+        mm.model_from_str('1.13.21')  # month>12
+
+
+def test_obj_processor_exception_wrap_for_common_rules():
+    grammar = r"""
+        Model: a+=A;
+        A: name=ID ('(' other=[A] ')')?;
+        """
+
+    @textxerror_wrap
+    def p(a):
+        if a.name == 'E':
+            raise Exception("test")
+
+    mm = metamodel_from_str(grammar)
+    mm.register_obj_processors({'A': p})
+    from textx.exceptions import TextXError
+    mm.model_from_str('X Y Z (X)')
+    with pytest.raises(TextXError, match=r'None:1:3:.*test'):
+        mm.model_from_str('X E Z (X)')
+
+
+def test_obj_processor_exception_wrap_for_match_rules():
+    grammar = r"""
+        Model: a+=A;
+        A: /\w+/;
+        """
+
+    @textxerror_wrap
+    def p(a):
+        if a == 'E':
+            raise Exception("test")
+
+    mm = metamodel_from_str(grammar)
+    mm.register_obj_processors({'A': p})
+    from textx.exceptions import TextXError
+    mm.model_from_str('X Y Z')
+    with pytest.raises(TextXError, match=r'None:1:3:.*test'):
+        mm.model_from_str('X E Z')
--- a/tests/functional/test_user_classes.py
+++ b/tests/functional/test_user_classes.py
@@ -95,10 +95,17 @@ def test_user_class_with_imported_grammar():
     this_folder = dirname(abspath(__file__))
     mm = metamodel_from_file(join(this_folder, "user_classes", "B.tx"),
                              classes=[AThing, BThing])
+    called = [False]
+
+    def dummy(_):
+        called[0] = True
+
+    mm.register_obj_processors({'AThing': dummy})
     m = mm.model_from_str("""
         A 2,1
         B Hello
     """)
+    assert called[0]
     assert m.a.v.x == 2
     assert m.a.v.y == 1
     assert m.b.v.name == "Hello"
--- a/textx/metamodel.py
+++ b/textx/metamodel.py
@@ -138,7 +138,7 @@ class TextXMetaModel(DebugPrinter):
             to use instead of the dynamically created or a callable providing
             those classes. The callable must accept a rule name and return a
             class for that rule name or None.
-        obj_processors(dict): A dict of user supplied object processors keyed
+        _obj_processors(dict): A dict of user supplied object processors keyed
             by rule/class name (may be a fully qualified name).
         rootcls(TextXClass): A language class that is a root of the meta-model.
         root_path(str): The root dir used for the import statement.
@@ -213,11 +213,8 @@ class TextXMetaModel(DebugPrinter):
         # Registered model processors
         self._model_processors = []
 
-        # Registered object processors
-        self.obj_processors = {}
-
         # Match rule and base type conversion callables
-        self.type_convertors = {
+        self._default_obj_processors = {
             'BOOL': lambda x: x == '1' or x.lower() == 'true',
             'INT': lambda x: int(x),
             'FLOAT': lambda x: float(x),
@@ -225,6 +222,9 @@ class TextXMetaModel(DebugPrinter):
             'STRING': lambda x: x[1:-1].replace(r'\"',
                                                 r'"').replace(r"\'", "'"),
         }
+
+        # Registered object processors (use _default_obj_processors)
+        self.register_obj_processors({})
 
         # Registered scope provider
         self.scope_providers = {}
@@ -498,11 +498,38 @@ class TextXMetaModel(DebugPrinter):
         clazz._tx_attrs[name] = attr
         return attr
 
-    def convert(self, value, _type):
+    def has_obj_processor(self, _type):
+        return _type in self._obj_processors
+
+    def process(self, value, _type, filename, col, line):
         """
+        Process a value with the given type
         Convert instances of textx types and match rules to python types.
+
+        Args:
+            value: the value
+            _type: the type of the value (TextX Class)
+            filename: filename for current object
+            col: col for current object
+            line: line for current object
+        Returns:
+            None or a value used by TextX to replace the object during
+            model creation.
         """
-        return self.type_convertors.get(_type, lambda x: x)(value)
+        try:
+            return self._obj_processors.get(_type, lambda x: x)(value)
+        except Exception as e:
+            from textx.exceptions import TextXError
+            if isinstance(e, TextXError):
+                if e.col is None:
+                    e.col = col
+                if e.line is None:
+                    e.line = line
+                if e.filename is None:
+                    e.filename = filename
+                raise e
+            else:
+                raise
 
     def validate(self):
         """
@@ -708,8 +735,8 @@ class TextXMetaModel(DebugPrinter):
             obj_processors(dict): A dictionary where key=class name,
                 value=callable
         """
-        self.obj_processors = obj_processors
-        self.type_convertors.update(obj_processors)
+        self._obj_processors = dict(self._default_obj_processors)
+        self._obj_processors.update(obj_processors)
 
     @property
     def _tx_model_param_definitions(self):
--- a/textx/model.py
+++ b/textx/model.py
@@ -5,9 +5,10 @@ Model construction from parse trees and the model API.
 import sys
 import codecs
 import traceback
+from future.utils import raise_from
 from collections import OrderedDict
 from arpeggio import Parser, Sequence, NoMatch, EOF, Terminal
-from textx.exceptions import TextXSyntaxError, TextXSemanticError
+from textx.exceptions import TextXError, TextXSyntaxError, TextXSemanticError
 from textx.const import MULT_OPTIONAL, MULT_ONE, MULT_ONEORMORE, \
     MULT_ZEROORMORE, RULE_ABSTRACT, RULE_MATCH, MULT_ASSIGN_ERROR, \
     UNKNOWN_OBJ_ERROR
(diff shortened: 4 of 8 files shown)


goto40:
@igordejanovic: to be discussed. I think this could be useful.

goto40:
We could use `raise from`, see https://stackoverflow.com/questions/24752395/python-raise-from-usage

However, it seems not to be python 2.7 compatible...

igordejanovic:
@goto40 That's a good idea but there are two problems with the approach IMO:
1. It is backward incompatible. I know for a project that depends on exceptions from object processors being propagated unchanged.
2. It is misleading to do implicit conversion of exceptions. It is something users won't expect.

I would like to discuss some alternative approaches. Like, e.g. providing a decorator which when applied to an object processor would yield this behavior (wrapping exception in `TextXError`), or alternatively might just add `line/col/filename` info to the original exception. I agree with usage of `raise from` in case of wrapping. This could be implemented starting with textX 3 as we drop support for Python 2.

goto40:
Good points! Let me try an alternative with decorators...

goto40:
> Like, e.g. providing a decorator which when applied to an object processor would yield this behavior (wrapping exception in `TextXError`), or alternatively might just add `line/col/filename` info to the original exception. I agree with usage of `raise from` in case of wrapping. This could be implemented starting with textX 3 as we drop support for Python 2.

Thank you very much for your comments. I tried a new version:

 * decorators (I added a decorator `object_processor`): now you have the choice, how to handle errors (default=as before)
 * in some cases no model element is present (when calling `convert`) - in that case a TextXError without position information is thrown, this TextXError is catched and empty fields are filled in (no exactly what you depicted - maybe a you see a better solution). Note: This case - I think - cannot be handled outside textx, because the model location is lost. We should carefully review that part if we vote the solution to be adequate.
 * `raise from`: I included it, but I did not play with it a lot - maybe the next days...

igordejanovic:
> in some cases no model element is present (when calling convert)

Maybe the decorator should just add a maker on the function and the real wrapping should be done where the info is available (if the marker is present).

BTW, not sure that decorator should be called `object_processor`. It implies that each processor should be decorated and might be confusing. Maybe something that will give a clue what the decorator is doing (e.g. `textxerror_wrap` ?).

goto40:
OK for the decorator name. Good point.

goto40:
For the marker: I think the decorator way (with the decorator doing the job) is nice and clear. The only problem is that sometimes the location is unknown. This is solved by detecting textxErrors without location (and adding the missing info). This - if I do not oversee something - will not get clearer by adding a marker... I will refactor the PR. Then we can continue discussing. Feel free to elaborate on the marker idea.

igordejanovic:
> This is solved by detecting textxErrors without location (and adding the missing info). This - if I do not oversee something - will not get clearer by adding a marker...

I agree. That solution is better.

goto40:
I refactored the code. If we agree I could continue with the documentation...
Question: did I catch all places, where we call an object processor without having a full object (with _tx_position)?

Comments welcome!

goto40:
(will check the python 2 issue this weekend)

igordejanovic:
@goto40 Overall looks good but I have several suggestions:

 - [x] We have `type_convertors` and `obj_processors` which seems to serve the same purpose. I would like to unify these and have only a single meta-model call for processing/converting. We may call that meta-model method `process` or `convert`, whatever (maybe `process` is better as conversion might not occur). Current `type_convertors` are default obj. processors which should be always there and the user registered would override/update:

```python
def register_obj_processors(obj_processors):
    self.obj_processors = dict(self.default_obj_processors)    # these default_obj_processors are current type_convertors
    self.obj_processors.update(obj_processors)
```

 - [x] Treat `obj_processors` in meta-model as private and rework/change all external references (e.g. in the `model.py`) to `process` call. Do the same for all `convert` calls.

 - [x] Now, we can put `filename, col, line` as a mandatory parameters to `metamodel.process` so we will get error if position information is not passed in.

 - [x] Do the textx error wrapper update in a single location, `metamodel.process` method.

I think these changes will make the code more clean If I didn't overlook something.

goto40:
@igordejanovic, just a quick question: Am I right, that all user classes have a "_tx_fqn" fields. It seems so (I explored interactively) - I do not remember exactly what we defined when implementing the attr-feature for user classes... (do we support user classes with slots?)

igordejanovic:
> @igordejanovic, just a quick question: Am I right, that all user classes have a "_tx_fqn" fields. It seems so (I explored interactively) - I do not remember exactly what we defined when implementing the attr-feature for user classes... (do we support user classes with slots?)

`_tx_fqn` is introduced [here](https://github.com/textx/textX/commit/971ecc6). User classes with frozen attributes is started [here](https://github.com/textx/textX/commit/4cf40e9). 

From what I can tell by investigating the code and tests with user classes we do provide `_tx_fqn` and other `_tx_` class-level attributes, even for user classes with slots.

textx/model.py
```diff
@@ -693,7 +743,12 @@ def call_obj_processors(metamodel, model_obj,
         # enter recursive visit of attributes only, if the class of the
         # object being processed is a meta class of the current meta model
         if model_obj.__class__.__name__ in metamodel:
-            current_metaclass_of_obj = metamodel[model_obj.__class__.__name__]
+            if hasattr(model_obj, '_tx_fqn'):
+                current_metaclass_of_obj = metamodel[model_obj._tx_fqn]
+            else:
+                # fallback (not used - unsure if this case is required...):
+                current_metaclass_of_obj = metamodel[model_obj.__class__.__name__]
+            assert current_metaclass_of_obj is not None
```
goto40, line 745: @igordejanovic This code block should be reviewed carefully. I had issues with grammars where rules are redefined (`test_user_class_with_imported_grammar` in https://github.com/textX/textX/blob/master/tests/functional/test_user_classes.py#L94) during code changes. I did never run into the fallback case - and (since we always provide a `_tx_fqn` field) we could probably remove the else case...

goto40:
Thank you for the comments.

Code changes done. Ready for next review step... (see inline comment above, for the most critical code change in my eyes)

tests/functional/test_processors.py
```diff
@@ -350,3 +350,60 @@ def myfloat_processor(x):
     assert called[0]
     assert model.objects[0] == '#3.4'
     assert model.objects[1] == '--6--'
+
+
+def test_exception_from_obj_processor_has_proper_type1():
```
igordejanovic: Maybe to put a more specific test name. Like `test_obj_processor_exception_wrap_for_external_exceptions`

tests/functional/test_processors.py
```diff
@@ -350,3 +350,60 @@ def myfloat_processor(x):
...
+
+    mm = metamodel_from_str(grammar)
+    mm.register_obj_processors({'Date': date_converter})
+    mm.model_from_str('1.12.20')  # ok
+    from textx import TextXError
+    with pytest.raises(TextXError, match=r'None:1:1:.*1\.\.12'):
+        mm.model_from_str('1.13.21')  # month>12
+
+
+def test_exception_from_obj_processor_has_proper_type2():
```
igordejanovic: This could be named `test_obj_processor_exception_wrap_for_common_rules`

tests/functional/test_processors.py
```diff
@@ -350,3 +350,60 @@ def myfloat_processor(x):
...
+
+    mm = metamodel_from_str(grammar)
+    mm.register_obj_processors({'A': p})
+    from textx.exceptions import TextXError
+    mm.model_from_str('X Y Z (X)')
+    with pytest.raises(TextXError, match=r'None:1:3:.*test'):
+        mm.model_from_str('X E Z (X)')
+
+
+def test_exception_from_obj_processor_has_proper_type3():
```
igordejanovic: This could be named `test_obj_processor_exception_wrap_for_match_rules`

tests/functional/test_user_classes.py
```diff
@@ -95,10 +95,17 @@ def test_user_class_with_imported_grammar():
...
+
+    def dummy(_):
+        called[0] = True
+
+    mm.register_obj_processors({'AThing': dummy})
     m = mm.model_from_str("""
         A 2,1
         B Hello
     """)
+    assert called[0]
```
igordejanovic, line 108: Is this directly related to this PR or just an enhancement of the existing test? If not it may still be here but squashed as a separate commit before merge.

textx/metamodel.py
```diff
@@ -213,19 +213,19 @@ def __init__(self, file_name=None, classes=None, builtins=None,
...
         self._model_processors = []
 
-        # Registered object processors
-        self.obj_processors = {}
-
         # Match rule and base type conversion callables
-        self.type_convertors = {
-            'BOOL': lambda x: x == '1' or x.lower() == 'true',
+        self._default_obj_processors = {
+            'BOOL': lambda x: x == '1' or x.lower() == 'true' or type(x) is bool,
```
igordejanovic: Why this `or type(x) is bool` extension?

textx/metamodel.py
```diff
@@ -498,11 +498,43 @@ def _new_cls_attr(self, clazz, name, cls=None, mult=MULT_ONE, cont=True,
         clazz._tx_attrs[name] = attr
         return attr
 
-    def convert(self, value, _type):
+    def has_obj_processor_for_type(self, _type):
```
igordejanovic: The name may be shorter - `has_obj_processor`

textx/model.py
```diff
@@ -185,6 +186,42 @@ def get_location(model_obj):
     return {"line": line, "col": col, "filename": the_model._tx_filename}
 
 
+def textxerror_wrap(obj_processor):
+    """
+    This is a decorator to wrap your object processors
+    if you desire to automatically convert any exception
+    into a TextXError (with file/position information).
+
+    This is useful to use non-textx exceptions
```
igordejanovic:
```suggestion
    This is useful when non-textx exceptions are used
```

textx/model.py
```diff
@@ -459,33 +496,46 @@ def process_match(nt):
         """
         Process subtree for match rules.
         """
+        line, col = parser.pos_to_linecol(nt.position)
         if isinstance(nt, Terminal):
-            return metamodel.convert(nt.value, nt.rule_name)
+            return metamodel.process(nt.value, nt.rule_name,
+                                     filename=parser.file_name,
+                                     line=line, col=col,
+                                     _default_processor=lambda x: x)
```
igordejanovic: I feel uneasy about `_default_processor` param. If we can figure out when this identity transformation should be used directly in the `process` method it would make these calls cleaner. If identity transformation is used always when there are no other processors two test cases are failing (related to calling processors in the context of abstract rules).

textx/model.py
igordejanovic: Yes, this might be a problem as the `if` condition is using a plain name as a guard to enter this block. I haven't investigated this in greater details as I want to provide the first feedback ASAP.

textx/model.py
```diff
@@ -718,16 +773,21 @@ def call_obj_processors(metamodel, model_obj,
...
-            metaclass_of_grammar_rule.__name__, None)
-        if obj_processor_grammar:
-            return_value_grammar = obj_processor_grammar(model_obj)
+        if metamodel.has_obj_processor_for_type(metaclass_of_grammar_rule.__name__):
+            try:
+                loc = get_location(model_obj)
+            except Exception:
+                raise TextXError("you may not define an object processor for "
+                                 "a class '{}' preventing to get the location "
+                                 "from.".format(metaclass_of_grammar_rule.__name__))
```
igordejanovic: This error message is not clear to me.

textx/model.py
```diff
@@ -843,11 +903,8 @@ def call_obj_processors(metamodel, model_obj,
 
                     # We have model loaded and all link resolved
                     # So we shall do a depth-first call of object
-                    # processors if any processor is defined.
-                    if m._tx_metamodel.obj_processors:
-                        if parser.debug:
-                            parser.dprint("CALLING OBJECT PROCESSORS")
-                        call_obj_processors(m._tx_metamodel, m)
+                    # processors
+                    call_obj_processors(m._tx_metamodel, m)
```
igordejanovic, line 899: This `dprint` should probably be now moved to `call_obj_processors`

igordejanovic (commented):
@goto40 Good work! It looks better. I have some more feedback. Please see bellow.

goto40:
Thank you for the comments. I will work on it.

tests/functional/test_processors.py
goto40: done

tests/functional/test_processors.py
goto40: done

tests/functional/test_processors.py
goto40: done

tests/functional/test_user_classes.py
goto40: ok. Good point. Added to TODO list above.

textx/metamodel.py
goto40: done

textx/metamodel.py
goto40: Good catch! This was an error. I had issues in a first version, where I called obj processors multiple times. However, the fix here was wrong. Reverted to original version.

textx/model.py
goto40: Initially I had not the "if obj_processor is defined" statement above - and always called the object_processor for the current element. It then happened, that the object processor was called for an "int" or "str" causing an error.

I do not think we run into that case with the additional if. So I removed the try-except-block (and the confusing error message).

Note: if the error happens, we get an exception anyhow... So no need to transform this error here!

textx/model.py
goto40: I added it outside the `call_obj_processors` (as before), since `call_obj_processors` is called recusively.

textx/model.py
goto40: My thoughts:
 * We could pass the type of rule. But this would be the same as passing a flag - or the function itself.
 * We could extract the information from the metamodel... I need to investigate how to do this - but it should be easy.
 * We could only **call process in the last stage of model creation only if an object processor is defined**. I try this - so we could discuss it. You are right - this seems maybe better. 

--> selected last version. updated code to be discussed.

textx/model.py
goto40: (code reactivated / reverted deletion)

igordejanovic:
@goto40 Sorry for the delay. It looks much better. Good work! Please proceed with completion of the PR.

igordejanovic (approved):

textx/model.py
goto40: @igordejanovic : Did you had time to investigate this point? If not we can either chose to (a) make more tests (I could do this the next time). Or (b) we can first merge and make this investigations separately (I think: if we find effects here, these effects are independent of this PR - and are just made visible by it...).

My favorite is to make the investigations separately... (to reduce the clutter induced by this PR getting too large).. What do you think?

goto40:
Added change log, squashed commits.

goto40:
Note: Travis seems to be fast again :-)

textx/model.py
igordejanovic: I agree. We can investigate that separately.

igordejanovic:
> Note: Travis seems to be fast again :-)

Yeah, it seem they solved some issues. It's good that we can work normally again. :)
Still, we should migrate due to their policy changes.

goto40:
@igordejanovic If you have no further comments I could merge this PR.

igordejanovic:
@goto40 No further comments. Go ahead and merge.

PR #320, opened 2021-08-03
Fix/improve error messages

igordejanovic:
I noticed that we have a problem with error messages returned when e.g. running the textx command. The problem was that when syntax error is encountered Arpeggio will return the exception where the location and the context is already baked into the exception string which made textX messages redundant.

E.g. previously we had:

```
(venv) ~/repos/textX/textX/examples/robot ±master⚡ » textx generate --grammar robot.tx program.rbt --target dot
   --overwrite
Generating dot target from models:
/home/igor/repos/textX/textX/examples/robot/program.rbt
Error: None:3:3: error: Expected 'initial' or 'up' or 'down' or 'left' or 'right' or 'end' at position
 /home/igor/repos/textX/textX/examples/robot/program.rbt:(3, 3) => 'al 3, 1   *north 4   '.

```
Notice that we have word `error`, filename and location printed twice. The filename is even wrong at the first occurrence (`None`).

This PR aims to fix this. But it required fix in the Arpeggio also. See the branch `nomatch-eval-attrs` in the Arpeggio repo.

## Code review checklist

### DIFF
--- a/tests/functional/regressions/test_issue188.py
+++ b/tests/functional/regressions/test_issue188.py
@@ -39,6 +39,6 @@ def test_issue188_skipws_5():
         SPACE[noskipws]: ' ';
     ''')
     with pytest.raises(TextXSyntaxError,
-                       match='.*Expected \'bar \' at position'):
+                       match=r'.*\d+: Expected \'bar \''):
         mm.model_from_str(' foo bar')
     mm.model_from_str(' foo bar ')
--- a/.github/workflows/ci-linux-ubuntu.yml
+++ b/.github/workflows/ci-linux-ubuntu.yml
@@ -23,6 +23,7 @@ jobs:
       run: |
         cd ..
         git clone --depth=1 --branch=master git://github.com/textX/Arpeggio.git
+        pip uninstall -y Arpeggio
         cd Arpeggio && python setup.py install
         cd ../textX
     - name: Run unit tests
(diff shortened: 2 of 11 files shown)


tests/functional/regressions/test_issue188.py
```diff
@@ -39,6 +39,6 @@ def test_issue188_skipws_5():
         SPACE[noskipws]: ' ';
     ''')
     with pytest.raises(TextXSyntaxError,
-                       match='.*Expected \'bar \' at position'):
+                       match='.*\d+: Expected \'bar \''):
```
goto40: seems a `\` is missing: `'.*\\d+: Expected \'bar \''`

goto40:
Yes. Error messages look sometimes a bit confusing. Very good!

Are you using a special editor (e.g. vscode) for which you optimize for? or just the command line? Even w/o syntax highlighting it should be easy to provide a plugin to allows parsing and/or generators to be integrated into such an IDE...

tests/functional/regressions/test_issue188.py
igordejanovic: Yup. good catch. Fixed.

igordejanovic:
> Are you using a special editor (e.g. vscode) for which you optimize for? or just the command line? Even w/o syntax highlighting it should be easy to provide a plugin to allows parsing and/or generators to be integrated into such an IDE...

I'm not using VS Code nor do I optimize for some specific editor. Yes, I agree, as long as editor can call external commands it should be easy to integrate parsing/code generation. I think that's what textX-LS is doing.

Issue #358, opened 2021-12-04
unordered group question


goto40:
According to http://textx.github.io/textX/stable/grammar/#repetitions an unordered group must be applied to a sequence.

However, in https://github.com/textX/textX/blob/master/examples/pyFlies/pyflies.tx#L168 it is applied to an ordered choice.

Maybe we should clarify this in the docs...? Or did I misunderstood something?

Once we clarify this (how it should be) I can make a proposition for the docs. I could also elaborate on how optionals are handled in unordered groups...

igordejanovic:
You are right. It wasn't clear enough. Actually, usage of ordered groups looks like an accidental feature :)
I did some more testing and addition to the docs to clarify. Will issue a PR shortly.

igordejanovic:
Closed via #359

PR #359, opened 2021-12-04
Tests and improved docs for unordered groups

igordejanovic:
refs #358

## Code review checklist

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -47,11 +47,11 @@ depicted in the following figure:
 * `INT` rule: matches an integer number. This match will be converted to
   a Python `int` instance.
 * `FLOAT` rule: will match a floating point number. This match will be converted
-  to a Python `float` instance ('FLOAT' is a direct subtype of 'BASETYPE'; 
+  to a Python `float` instance ('FLOAT' is a direct subtype of 'BASETYPE';
   in order to distinguish floats and ints, 'STRICTFLOAT' was introduced).
-* `STRICTFLOAT` rule: will match a floating point number. This match will be 
-  converted to a Python `float` instance. A 'STRICTFLOAT' will not match an 'INT' 
-  (without "." or "e|E"). A 'NUMBER' is either a 'STRICTFLOAT' or an 'INT', 
+* `STRICTFLOAT` rule: will match a floating point number. This match will be
+  converted to a Python `float` instance. A 'STRICTFLOAT' will not match an 'INT'
+  (without "." or "e|E"). A 'NUMBER' is either a 'STRICTFLOAT' or an 'INT',
   and will, thus, be converted to a float or an int, respectively.
 * `BOOL` rule: matches the words `true` or `false`. This match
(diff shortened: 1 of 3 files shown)


docs/grammar.md
```diff
@@ -262,8 +262,8 @@ Now, an ordered choice in the parentheses is optional.
 
 * **Unordered group** is a special kind of a sequence. Syntactically it is
   similar to a repetition. It is specified by the `#` operator and must be
-  applied to sequences. This operator will match each element of the sequence in
-  an arbitrary order:
+  applied to either sequences or ordered choices. This operator will match each
+  element of the sequence in an arbitrary order:
```
goto40: ...will match each element of the sequence +"or the ordered choice"  in an arbitrary order:

goto40:
very good. The addition about ordered choices makes the thing clear.

One more thing: We could add more infos on optional parts of the unordered group:
 * optional parts (e.g. `x ?= 'A'`  work as you intuitively expect: they are accepted in an arbitrary order
 * optional parts are optional - but if all of them are missing and there is no non-optional part, we have a miss (is this correct?)

I am playing with an experimental for-fun-version of parts of textx in c++. There I learn a lot about the internals of textx/arpeggio. I added a unittest on optional parts of unordered groups as I understood it (from the python code): https://github.com/goto40/textx-cpp/blob/master/test/arpeggio.t.cpp#L399

examples/pyFlies/pyflies.tx
```diff
@@ -165,7 +165,7 @@ Block:
 ;
 
 TestInstance:
-  'test' type=[TestType] trials=INT (practice?="practice"|randomize?="randomize")#
+  'test' type=[TestType] trials=INT (practice?="practice" randomize?="randomize")#
```
goto40, line 168: "Like"

goto40 (approved):
LGTM

igordejanovic:
@goto40 
> optional parts are optional - but if all of them are missing and there is no non-optional part, we have a miss (is this correct?)

No, it works as expected. If all elements are optional then even if there is no match of any of the optional parts the whole match still succeeds. At least that's how it should work and a quick test confirms it. The only issue I noticed is that if the whole unordered match is contained in a single common rule than a reference to that rule will return `None` in case all optionals are not matched. Is this "the correct" behavior I don't know. It can be either this or to return an object where all attributes are None. We can investigate this as a separate issue if needed.

> I am playing with an experimental for-fun-version of parts of textx in c++. There I learn a lot about the internals of textx/arpeggio. I added a unittest on optional parts of unordered groups as I understood it (from the python code): https://github.com/goto40/textx-cpp/blob/master/test/arpeggio.t.cpp#L399

Hey, that is excellent news! I look forward for the c++ implementation of textX :)

igordejanovic:
@goto40 Just to clear a bit on my previous statement 

> if there is no match of any of the optional parts the whole match still succeeds

What I mean is that the whole match still succeeds as it is considered optional as a whole.

Issue #366, opened 2022-01-30
Possible (minor) issue with attribute types 


goto40:
Playing around with the meta model properties I added some tests: https://github.com/textX/textX/tree/feature/additional_metamodel_tests

I found a possible issue I would like to discuss:

In the docu we have **"If there are multiple common rules than the first will be used as a result and the rest only for parsing"**.

Based on this sentence I tried:
```
        Model: 'value' x=Base;
        Base: S1|(S2 NotS3); // only use first rule (S2) as possible instance for each choice-option
        S1: name=ID;
        S2: name=ID;
        NotS3: name=ID;
```
In this meta model `Base` should not be a base of `NotS3`... I think this is a possible issue (probably not so difficult to fix).

igordejanovic:
I think you are right. It looks like a potential problem. The metamodel should be consistent with the parsing results. Since `NotS3` instance is never instantiated during `Base` parsing (i.e. it will be thrown away after parsing) it shouldn't be its subclass.

goto40:
I will have a look...

PR #369, opened 2022-02-16
Feature/additional metamodel tests

goto40:
See #366: I identified an issue with the meta model not indicating the correct inherited-by information.
I debugged the code and matched it with my understanding of the textx-grammar. 
 * I found that there was no mimic of stopping after the first non-match-rule (as described in https://textx.github.io/textX/stable/grammar/: "If there are multiple common rules than the first will be used as a result and the rest only for parsing") 
 * Based on the insights i added a simple fix by stopping the insertion process in `lang.py`.

This solution is quite simple. We should have a deep look, @igordejanovic. In case of doubts I could add more "negative tests" if we sketch them in this discussion... 
 
## Code review checklist

- [n/a] Commit messages are meaningful (see [this][commit messages] for details) - summary comment will be generated
- [X] Standalone docs have been updated accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,12 +41,14 @@ please take a look at related PRs and issues and see if the change affects you.
 
 ### Changed
 
+- Inheritance chain calculation. Possible **(BIC)** ([#369]).
 - Added `def_file_name` attribute to `RefRulePosition` for storing the definition's
   model file name in case of cross-references between models. ([#313],[#277])
 - Migrated from Travis CI to GitHub Actions ([#307])
 - Dropped support for deprecated Python versions. The lowest supported version
   is 3.6. **(BIC)**
 
+[#369]: https://github.com/textX/textX/pull/369
 [#360]: https://github.com/textX/textX/issues/360
 [#323]: https://github.com/textX/textX/issues/323
 [#320]: https://github.com/textX/textX/pull/320
--- a/tests/functional/test_metamodel_properties.py
+++ b/tests/functional/test_metamodel_properties.py
@@ -1,0 +1,94 @@
+from __future__ import unicode_literals
+import textx
+
+
+def test_metamodel_rule_properties():
+    mm = textx.metamodel_from_str(r'''
+        Model: 'value' a=C b=K c=N d=O e=P q=Q l=L;
+        C: D|E;
+        D: 'D';
+        E: F;
+        F: 'F';
+        K: L|M;
+        L: 'L' name=ID;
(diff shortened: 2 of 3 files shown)


goto40:
@igordejanovic : in particular, I am not sure about where the else part of `if type(rule) is OrderedChoice:` kicks in (just below the modified code in `lang.py`.

igordejanovic:
:+1: Will take a look in the next few days.

igordejanovic:
@goto40 This seems good. I took an opportunity to do some code simplification at this place. Please see that last commit.

igordejanovic:
I was thinking about some use-cases and have to verify before merging.

goto40:
Take your time... the change is minimal and will not cause merge conflicts. So we can let it open for a while... I stumbled over the bug while implementing the C++ version: https://github.com/goto40/textx-cpp/blob/master/test/metamodel.t.cpp#L571.

Re-implementing allowed me to get some deeper insights...

igordejanovic:
Nice progress with textx-cpp. Looks cool. :)

igordejanovic:
I've added another interesting case to the tests and fixed the implementation.

goto40:
Good catch...

goto40:
We should wait a little bit before merging... I am not sure if we have all border cases...

goto40:
Should we also talk about unordered groups? I think those should be handled like choice nodes... Each alternative can determine the output type. What do you think?

Apart this I am very happy...

goto40:
Hmm... Also optional parts need to be processed correctly... And what happens with zero-or-more?Maybe we leave this pr as it is and create a separate issue? In that case it looks good to me.

igordejanovic:
Good questions. We may leave this open for a while to add more tests. My initial hunch is that all repetitions should be treated as the first element of a sequence and only the first level of sequence (a sequence that is directly a part of ordered choice alternative) is processed in such a way to take only the first element for inheritance calculation. But, as you suggested, optional match may change the logic.

For example:
```
A: B | C D;      // only C inherits A but not D
A: B | C? D;     // Both C and D should inherit A as we may have D instantiated if C is not found
A: B | (C D)#    // Both C and D should inherit A as the order is changing
```

Since these are contrived, although possible usages, we should probably think a bit more before merging.

goto40:
OK.

igordejanovic:
@goto40 We can provide additional improvement in new PRs linked to this and #366

goto40:
> @goto40 We can provide additional improvement in new PRs linked to this and #366

Ok

PR #440, opened 2026-07-04
fix: multiplicity inference for attributes with multiple assignments

### LINKED ISSUE #357: Possible inconsistent multiplicity of attributes
While playing with the code I found an effect concerning the multiplicity of attributes in case of multiple assignment to the same attribute (http://textx.github.io/textX/stable/grammar/#multiple-assignment-to-the-same-attribute) in one sequence or in more complicated parse tree constellations (see "OK"/"FAILED" below).

```
def test_attribute_multiplicity():
    """
    Test http://textx.github.io/textX/stable/grammar/#multiple-assignment-to-the-same-attribute
    """
    grammar = """
    Rule1: a='a' a='c';
    Rule2: (a='a'|a='b') a='c';
    """
    meta = metamodel_from_str(grammar)
    assert meta['Rule1']._tx_attrs['a'].mult == '1..*'            // OK
    assert meta['Rule2']._tx_attrs['a'].mult == '1..*'            // FAIL (result="1")
```

Is it worth to make further investigations? Or is this a border case and we can adapt the doc? Or did I misunderstood something?

mayank-dev-15:
Fixes #357

When an attribute appears in multiple assignments in an ordered choice followed by another assignment in the parent sequence (e.g. `(a='a'|a='b') a='c'`), the multiplicity was incorrectly inferred as "1" instead of "1..*" because assignments inside ordered choice branches were not propagated back to the parent scope.

Fix: Union all branch attribute sets after processing an OrderedChoice and propagate them to the parent scope so duplicate assignment detection works across the entire sequence.

### DIFF
--- a/tests/functional/regressions/test_issue357.py
+++ b/tests/functional/regressions/test_issue357.py
@@ -1,0 +1,88 @@
+"""
+Tests for issue: https://github.com/textX/textX/issues/357
+Inconsistent multiplicity of attributes with multiple assignment.
+"""
+
+from textx import metamodel_from_str
+
+
+def test_multiplicity_with_multiple_assignments_in_ordered_choice():
+    """
+    When an attribute appears in multiple assignments in an ordered choice
+    followed by another assignment in the parent sequence, the multiplicity
+    should be "1..*".
+    Grammar: (a='a'|a='b') a='c'
+    """
+    grammar = r"""
+    Model: (a='a' | a='b') a='c' 'end';
+    """
+    mm = metamodel_from_str(grammar)
+    attr = mm["Model"]._tx_attrs["a"]
+    assert attr.mult == "1..*", (
+        f"Expected multiplicity '1..*' for attribute 'a' with multiple "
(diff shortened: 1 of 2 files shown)


tests/functional/regressions/test_issue357.py
```diff
@@ -0,0 +1,88 @@
...
+    and also in a sequential assignment, multiplicity should be "1..*".
+    """
+    grammar = r"""
+    Model: (a='x' | a='y' | a='z') a='w' 'end';
+    """
+    mm = metamodel_from_str(grammar)
+    attr = mm["Model"]._tx_attrs["a"]
+    assert attr.mult == "1..*", (
+        f"Expected multiplicity '1..*' for attribute 'a', got '{attr.mult}'"
+    )
```
igordejanovic, line 39: These two tests above are testing the same thing. Just one should remain.

igordejanovic:
Thanks @mayank-dev-15 . Please, also format the core with `just format`

igordejanovic (changes requested):

PR #444, opened 2026-07-18
fix: export falsy empty-string match models

santhreal:
model_export_to_file hits Exception when falsy empty-string match model rejected by 'if not model'. This PR fixes the regression with a focused test covering the case. ## Notes (awaiting thermo)

### DIFF
--- a/tests/functional/regressions/test_export_falsy_match_model.py
+++ b/tests/functional/regressions/test_export_falsy_match_model.py
@@ -1,0 +1,25 @@
+import io
+
+from textx import metamodel_from_str
+from textx.export import model_export_to_file
+
+
+def test_export_empty_match_rule_model():
+    """Match-rule roots may be empty strings; export must not treat them as omitted."""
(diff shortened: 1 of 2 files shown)


tests/functional/regressions/test_export_falsy_match_model.py
```diff
@@ -0,0 +1,25 @@
...
+    text = out.getvalue()
+    assert "digraph textX" in text
+
+
+def test_export_rejects_missing_model_and_repo():
+    out = io.StringIO()
+    try:
+        model_export_to_file(out, None, None)
+        assert False, "expected Exception"
+    except Exception as err:
```
igordejanovic, line 24: Please use `with pytest.raises...` instead of try/except. Also be sure to run `just check`.

igordejanovic (changes requested):
Thanks @santhreal for the contribution. Please see inline comment. Also, please state if this change is AI assisted for transparency.

Issue #59, opened 2018-03-19
Incorrect edge case in treating escaped characters in quoted strings


jhrmnn:
I believe that the following is parsed incorrectly:

```
Python 3.6.4 (default, Mar  9 2018, 23:15:03)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(ins)>>> from textx.metamodel import metamodel_from_str
(ins)>>> mm = metamodel_from_str('Root: STRING;')
(ins)>>> mm.model_from_str(r"'\\' a'")
"\\' a"
```

In my understanding, `\\` should be interpreted as a single backslash, the following `'` should terminate the string, which should leave the hanging ` a'` as non-parseable. At least that's what Python does.

jhrmnn:
Er, sorry, this should have gone to Arpeggio, I guess.

igordejanovic:
No problem. :)

Yes, It seems that you are right. I'll have to do some more investigation.

PR #445, opened 2026-07-19
Only unescape the delimiting quote in a STRING

chuenchen309:
The default `STRING` base-type object processor unescapes both `\"` and `\'`
regardless of which quote delimits the string:

```python
"STRING": lambda x: x[1:-1].replace(r'\"', '"').replace(r"\'", "'"),
```

But the `STRING` match regex only treats the *delimiting* quote as escapable
(`"(\"|[^"])*"` / `'(\'|[^'])*'`). Inside a double-quoted string a backslash
before a single quote is a literal backslash, not an escape — so the processor
silently deletes it:

```python
mm = metamodel_from_str('Model: s=STRING;')
mm.model_from_str(r'''"caf\'e"''').s   # -> 'cafe'   (the \ is dropped; expected: r"caf\'e")
mm.model_from_str(r"""'a\"b'""").s     # -> 'a"b'    (expected: r'a\"b')
```

Unescape only the delimiter, chosen from the opening quote. The existing
`test_string_escaping` already pins the quote-specific semantics (embedded
non-delimiter quotes stay literal); this extends it with a backslash before the
non-delimiting quote. Escaping the delimiter itself (`"a\"b"` -> `a"b`) is
unchanged, and the core suite still passes.

This is distinct from #59 (which is about collapsing `\\`).

---
This PR was authored by an AI coding agent (Claude Code) running on this account:
the AI found the bug, ran the repro, wrote the test, and wrote this description.
The human account holder reviews every change and is accountable for it. The
verification is real and re-runnable from the diff. If this isn't the kind of
contribution you want, say so and I'll close it.

### DIFF
--- a/tests/functional/test_textx_language.py
+++ b/tests/functional/test_textx_language.py
@@ -481,7 +481,9 @@ def test_string_escaping():
     "Double quotes string", "Double quotes with 'single quotes embedded'",
     "Double quotes with \" escaped quotes",
     'Single quotes string', 'Single quotes with "double quotes embedded"',
-    'Single quotes with \' escaped single quotes'
+    'Single quotes with \' escaped single quotes',
+    "Double quotes with literal backslash \' quote",
+    'Single quotes with literal backslash \" quote'
     """
     )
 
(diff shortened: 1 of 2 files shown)


igordejanovic:
Thanks @chuenchen309 for the contribution and being transparent about AI assistance.

